PR #1, opened 2025-09-29
feat: Add GitHub workflows for automated Helm chart testing and validation

jordigilh:
- Add comprehensive deployment test workflow (test-deployment.yml):
  * Runs deploy-kind.sh and install-helm-chart.sh scripts
  * Tests complete E2E deployment on PRs and merges
  * Includes health checks, connectivity tests, and cleanup
  * 45-minute timeout with robust error handling and diagnostics

- Add simplified Helm validation workflow (lint-and-validate.yml):
  * Uses helm lint and helm template --validate --strict
  * Fast 10-minute validation for chart correctness
  * Tests template generation and Kubernetes schema compliance
  * Handles dependency checking if present

Both workflows trigger on PR/push to main/master branches with path filtering
for ros-ocp/** changes, ensuring automated testing of chart deployments.

### DIFF
--- a/.github/workflows/lint-and-validate.yml
+++ b/.github/workflows/lint-and-validate.yml
@@ -1,0 +1,45 @@
+name: Lint and Validate Helm Chart
+
+on:
+  pull_request:
+    branches: [ main, master ]
+    paths:
+      - 'ros-ocp/**'
+  push:
+    branches: [ main, master ]
+    paths:
+      - 'ros-ocp/**'
+  workflow_dispatch:
+
+jobs:
+  helm-validation:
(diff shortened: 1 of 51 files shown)


jordigilh:
@gciavarrini PTAL

jordigilh:
@masayag @ydayagi FYI. Since the chart is required in the new ingress and ocp-backend as part of the smoke test validation, I'm extracting it from the ros-ocp-backend to its own repo. Then I will update both new ingress and ocp-backend to pull the tgz from it rather than use the local copy. This way we can still maintain a smoke test for both repos without having to duplicate the chart

PR #4, opened 2025-09-30
Add Valkey support for OpenShift deployments

gciavarrini:
Enable platform-aware cache deployment using Valkey on OpenShift and Redis on Kubernetes.

Update service discovery to dynamically connect to the appropriate cache backend based on platform detection.

### DIFF
--- a/ros-ocp/templates/deployment-redis.yaml
+++ b/ros-ocp/templates/deployment-redis.yaml
@@ -1,47 +1,73 @@
+{{- $isOpenShift := eq (include "ros-ocp.isOpenShift" .) "true" }}
+{{- $cacheName := ternary "valkey" "redis" $isOpenShift }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ include "ros-ocp.fullname" . }}-redis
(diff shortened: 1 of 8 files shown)


ros-ocp/values.yaml
```diff
@@ -97,6 +97,15 @@ redis:
...
   port: 6379
 
+# Valkey (Production Cache - for OpenShift deployments)
+valkey:
+  image:
+    # Primary image (Red Hat)
+    repository: registry.redhat.io/rhel10/valkey-8
+    tag: "latest"
+  bindAddress: "0.0.0.0"
+  port: 6379
```
jordigilh, line 107: same port, do you think we can hardcode it?

ros-ocp/templates/deployment-redis.yaml
```diff
@@ -50,7 +76,11 @@ spec:
           readinessProbe:
             exec:
               command:
+                {{- if $isOpenShift }}
```
jordigilh, line 79: can you extract this into a helper function and capture the value at the beginning of the template instead of calling this logic again and again?

gciavarrini: wdyt?

PR #5, opened 2025-10-01
Add ODF support for production OpenShift deployments

gciavarrini:
Implement dual storage support in the ROS-OCP Helm chart, allowing `MinIO` for development (Kubernetes/KinD) and `ODF` for production (OpenShift) environments.

Key changes include:
- Added `ODF` configuration section to `values.yaml` with S3-compatible settings
- Created storage abstraction helpers in `_helpers.tpl` for seamless switching
- Updated deployment templates to conditionally deploy `MinIO` only on Kubernetes
- Enhanced `install-helm-chart.sh with automatic storage type detection
- Updated README.md with comprehensive storage configuration documentation

The solution maintains API compatibility between MinIO and ODF through their shared S3-compatible interface, ensuring minimal changes to application code while providing enterprise-grade storage for production deployments.


ros-ocp/templates/deployment-ingress.yaml
```diff
@@ -38,7 +39,19 @@ spec:
                 sleep 5
               done
               echo "MinIO is ready"
-              echo "MinIO is ready"
+        {{- else }}
+        - name: wait-for-odf
```
jordigilh, line 43: why not keep the same container and just change the command? Instead of having `wait-for-minio` and `wait-for-odf`. Once you start adding other types you'll end up with many more potential containers.. it's just for simplicity that the container that waits for the storage to be ready is always there with the same name

ros-ocp/templates/ingress.yaml
```diff
@@ -69,13 +69,15 @@ spec:
               name: {{ include "ros-ocp.fullname" $ }}-sources-api
               port:
                 number: {{ $.Values.sourcesApi.port }}
-        # MinIO Console (object storage UI)
+        # MinIO Console (object storage UI) - only for Kubernetes
+        {{- if not (eq (include "ros-ocp.isOpenShift" $) "true") }}
```
jordigilh, line 73: again, what do you think of just testing for "false" and remove the `not`?

ros-ocp/values.yaml
```diff
@@ -121,6 +121,26 @@ minio:
...
 
+# ODF (Production Storage - for OpenShift deployments)
+odf:
+  # ODF S3 endpoint (will be auto-detected from ODF installation)
+  endpoint: ""
+  # ODF S3 credentials (configure these values for your ODF installation)
+  # Get access key: oc get secret noobaa-admin -n openshift-storage -o jsonpath='{.data.AWS_ACCESS_KEY_ID}' | base64 -d
+  # Get secret key: oc get secret noobaa-admin -n openshift-storage -o jsonpath='{.data.AWS_SECRET_ACCESS_KEY}' | base64 -d
+  accessKey: "OUOJaGX15HKZ7XJ0imdT"
+  secretKey: "LYOBsoS5iH8lpyweJDjXUYWOgLwDA8Z2q/HG8rgu"
```
jordigilh, line 132: shouldn't this be somewhere outside the values.yaml? like in an existing secret that needs to be populated before the chart is created?

scripts/install-helm-chart.sh
```diff
@@ -167,8 +167,8 @@ download_latest_chart() {
 
     # Extract the tag name and download URL for the .tgz file
     local tag_name=$(echo "$latest_release" | jq -r '.tag_name')
-    local download_url=$(echo "$latest_release" | jq -r '.assets[] | select(.name | endswith(".tgz")) | .browser_download_url')
-    local filename=$(echo "$latest_release" | jq -r '.assets[] | select(.name | endswith(".tgz")) | .name')
+    local download_url=$(echo "$latest_release" | jq -r '.assets[] | select(.name | contains("latest")) | .browser_download_url')
```
jordigilh, line 170: thank you for fixing it! :)

ros-ocp/templates/secret-storage-credentials.yaml
```diff
@@ -1,9 +1,8 @@
-{{- $isOpenShift := eq (include "ros-ocp.isOpenShift" .) "true" }}
-{{- if not $isOpenShift }}
+{{- if eq (include "ros-ocp.isOpenShift" .) "false" }}
```
jordigilh, line 1: I'd say that instead of copying the admin credentials, the user must define the location of the secret in the namespace it can use instead. This means the namespace must exist first, but then we now make sure that we are not responsible for the credentials anymore (using the admin might not be safe and also the user might not have access to the namespace). It is probable that the secret is created via gitops with Vault/SecretManagement before the chart is created. WDYT?

PR #12, opened 2025-10-13
FLPATH-2682: Use Strimzi operator for Kafka instead of Helm templates

gciavarrini:
Refactor Kafka deployment by moving it out of the Helm chart and into the installation script using the Strimzi operator.

The Helm chart now focuses purely on application services, while the installation script handles Strimzi operator (`v0.45.1`) and Kafka cluster lifecycle.

The installation script auto-detects the platform (OpenShift vs Kubernetes), can reuse existing compatible Strimzi operators, and supports connecting to existing Kafka infrastructure via environment variables.

Configuration is cleaner with OpenShift-specific settings moved to `openshift-values.yaml`.

**Breaking changes:**
- Removed `strimzi.*` values from Helm chart
- No `ocp` argument (auto-detection instead)
- Kafka deployed in `kafka` namespace by script

**Compatibility:**
- Requires `Strimzi 0.43.x-0.45.x` for existing infrastructure
- Kafka `3.8.0` required

Signed-off-by: Gloria Ciavarrini <gciavarrini@redhat.com>

### DIFF
--- a/.github/workflows/test-deployment.yml
+++ b/.github/workflows/test-deployment.yml
@@ -1,6 +1,11 @@
 name: Helm Chart Quality Test
 
 # This workflow tests the ROS-OCP Helm chart deployment on Kind using the local chart
+# Deployment flow:
+#   1. Setup KIND cluster
+#   2. Deploy Kafka infrastructure (Strimzi + Kafka) - prerequisite
+#   3. Deploy ROS-OCP application
+#
 # Uses alternative registries to avoid Docker Hub rate limiting:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,12 @@ Kubernetes Helm chart for deploying the complete Resource Optimization Service (
 # Install latest release from GitHub
 ./scripts/install-helm-chart.sh
 
+# Or use local chart for development
```
jordigilh: Could it be that there should be a step before installing the helm chart where the `deploy-strimzi.sh` script is called? WDYT?

PR #19, opened 2025-10-27
FLPATH-2795: Fix ANSI color rendering in setup-cost-mgmt-tls

gciavarrini:
## Problem
The completion summary was displaying literal escape codes instead of colored output.

## Solution
Changed show_completion_summary() to use `echo -e` statements, consistent with other print functions in the script.

## Fixes 
https://issues.redhat.com/browse/FLPATH-2795


jordigilh (approved):
lgtm

PR #26, opened 2025-11-05
Fix Keycloak admin console 403, Failed to fetch, and Mixed Content errors

jordigilh:
## Problem

The Keycloak admin console was completely inaccessible due to three cascading issues:

1. **403 Forbidden** on `login-status-iframe.html` - prevented initial loading
2. **Failed to fetch TypeError** - occurred after successful login
3. **Mixed Content errors** - browser blocked HTTP requests on HTTPS page

### Issue 1: Missing CORS Configuration
The `security-admin-console` client had only relative paths in `redirectUris` and relied on `webOrigins: ["+"]`, which didn't work correctly with RHBK v22+ strict hostname validation.

### Issue 2: Strict Hostname Mode
Keycloak was configured with `hostname.strict: true` but missing the `hostname.admin` parameter, causing the admin console frontend to fail communicating with the backend API.

### Issue 3: Wrong Proxy Mode
`KC_PROXY=passthrough` told Keycloak that TLS is end-to-end, causing it to generate HTTP URLs since it received HTTP from the OpenShift router (which terminates TLS at the edge).

#### 1. Hostname Configuration
```yaml
hostname:
  hostname: $KEYCLOAK_HOSTNAME
  admin: $KEYCLOAK_HOSTNAME      # Added: explicit admin URL
  strict: false                    # Changed from true
  strictBackchannel: false
```

#### 2. Proxy Configuration
```yaml
unsupported:
  podTemplate:
    spec:
      containers:
      - name: keycloak
        env:
        - name: KC_PROXY
          value: edge                # Added: tells Keycloak TLS is terminated at router
```

#### 3. New Function: `configure_admin_console()`
Added automated configuration of the `security-admin-console` client via Keycloak Admin API:
- Sets explicit Web Origins: `https://<keycloak-url>`
- Sets explicit Redirect URIs: `https://<keycloak-url>/admin/master/console/*`
- Runs automatically during deployment after realm creation

## Testing

- ✅ Verified on OpenShift 4.x cluster with edge TLS termination
- ✅ Admin console loads without 403 errors
- ✅ Login succeeds without Failed to fetch errors
- ✅ All resources load over HTTPS (no Mixed Content errors)
- ✅ Full admin console functionality confirmed

## Impact

- **Existing deployments**: Will need manual patching (can apply same changes to existing Keycloak CR)
- **New deployments**: Will work out of the box with the updated script

## Related Issues

Fixes admin console access issues in RHBK v22+ deployments on OpenShift with edge TLS termination.

### DIFF
--- a/scripts/deploy-rhbk.sh
+++ b/scripts/deploy-rhbk.sh
@@ -396,7 +396,8 @@ spec:
     httpEnabled: true
   hostname:
     hostname: $KEYCLOAK_HOSTNAME
-    strict: true
+    admin: $KEYCLOAK_HOSTNAME
+    strict: false
     strictBackchannel: false
   ingress:
     enabled: true
@@ -404,14 +405,17 @@ spec:
     podTemplate:
       spec:
         containers:
-          - name: keycloak
-            resources:
-              requests:
-                cpu: 500m
-                memory: 1Gi
-              limits:
-                cpu: 1000m
-                memory: 2Gi
+        - name: keycloak
+          env:
+          - name: KC_PROXY
+            value: edge
+          resources:
+            requests:
+              cpu: 500m
+              memory: 1Gi
+            limits:
+              cpu: 1000m
+              memory: 2Gi
 EOF
         echo_success "✓ Keycloak instance created"
     fi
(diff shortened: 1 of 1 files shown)


jordigilh:
@masayag PTAL, this fixes a problem that @ELK4N4 reported to me where the keycloak console was not working correctly (not loading, just showing an image of spinning but nothing else). When I fixed that problem I noticed another one that occurred after logging in. Now I'm able to see the console.

masayag (approved):

PR #47, opened 2025-12-17
Add GH action to detect component updates

testetson22:
## Component Update Detection

Added automated monitoring for container image updates in the Helm chart.

### New Files

- **`.github/workflows/check-components.yml`** - GitHub Action that runs every 30 minutes to detect when `latest` tagged images have been updated. Creates an issue when changes are detected.

- **`scripts/qe/check-components.sh`** - Script that queries quay.io for digest changes on `latest`-tagged images. Used by the workflow and can be run locally.

- **`scripts/qe/test-gh-workflow-locally.sh`** - Helper to test GitHub workflows locally using `act`. Installs dependencies and handles `GITHUB_TOKEN`.

- **`scripts/qe/README.md`** - Documentation for QE scripts.

### How It Works

1. Parses `values.yaml` for images tagged as `latest` (pinned versions are ignored)
2. Queries quay.io API for current manifest digests
3. Compares against cached digests from previous runs
4. If changes detected: checks for an existing open issue with the component-update label, if found → updates the issue body with latest info, if not found → creates a new issue

### Usage

```bash
# Run workflow locally
./scripts/qe/test-gh-workflow-locally.sh

# List current component versions
./scripts/qe/test-gh-workflow-locally.sh -- --input mode=list-versions

# Run script directly
MODE=list-versions ./scripts/qe/check-components.sh
```

We should consider using renovate and/or dependabot to really manage dependencies, but my goal here is to just raise an issue if a component of the chart (that isn't version pinned) has changed then make a decision on when/if we need to trigger tests.

### DIFF
--- a/.github/workflows/check-components.yml
+++ b/.github/workflows/check-components.yml
@@ -14,7 +14,7 @@ name: Check Components
 
 on:
   schedule:
-    - cron: '*/30 * * * *'  # Every 30 minutes
+    - cron: '0 */4 * * *'  # Every 4 hours
   workflow_dispatch:
     inputs:
       mode:


chadcrum (approved):

.github/workflows/check-components.yml
```diff
@@ -0,0 +1,124 @@
...
+#   - list-versions: Report current digests and last modified dates
+#
+# Schedule: Every 30 minutes
+# Manual: Run via workflow_dispatch with mode selection
+
+name: Check Components
+
+on:
+  schedule:
+    - cron: '*/30 * * * *'  # Every 30 minutes
```
masayag: this might be too frequent. perhaps hourly or even every 4 hours.
otherwise, we might end up with having too many chart versions to verfiy.
on the other hand, having this frequent, will help us identify the offending image change when it will be detected. 

WDYT?

scripts/qe/check-components.sh
```diff
@@ -0,0 +1,121 @@
...
+has_updates="false"
+digests_json="{"
+first_digest=true
+
+while IFS= read -r image; do
+    [[ -z "$image" ]] && continue
+    
+    # Only check quay.io images
+    if [[ "$image" == quay.io/* ]]; then
+        repo_path="${image#quay.io/}"
```
masayag: In the future, this is going to be trickier when we'll move to registry.redhat.io

testetson22: Yea. I think we might eventually switch to the more robust dependabot and/or renovate configuration to actually manage/track dependencies. We could use the cut over to registry.redhat.io as the marker to switch to the more complex solution.

masayag (approved):

PR #84, opened 2026-02-08
Remove ingress RBAC and dedicated service account

masayag:
- Delete ClusterRole and ClusterRoleBinding for ingress (TokenReview unused when gateway handles JWT and injects X-Rh-Identity)
- Remove INGRESS_AUTH env from ingress deployment
- Remove dedicated ingress ServiceAccount; use namespace default
- Drop ingress.serviceAccount from values and doc INGRESS_AUTH reference


cost-onprem/templates/ingress/deployment.yaml
```diff
@@ -87,11 +86,6 @@ spec:
             - name: INGRESS_KAFKAGROUPID
               value: {{ .Values.ingress.kafka.groupId | default "ingress" | quote }}
 
-            # Authentication configuration
-            # JWT is handled by centralized gateway, backend trusts X-Rh-Identity header
-            - name: INGRESS_AUTH
-              value: "false"  # Auth is handled by gateway, backend trusts X-Rh-Identity
```
ELK4N4: Why we need auth if we use the gateway auth?

ELK4N4: The flag was false. It means the auth is enabled now

masayag: there is no use for that env-var.

https://github.com/search?q=org%3ARedHatInsights%20INGRESS_AUTH&type=code

masayag: TBH, even in ros-ingress we didn't use it.

ELK4N4: I'm not sure there is no usage in this var.
https://github.com/search?q=repo%3ARedHatInsights%2Finsights-ingress-go+options.SetEnvPrefix%28%22INGRESS%22%29&type=code

https://github.com/search?q=repo%3ARedHatInsights%2Finsights-ingress-go+options.SetDefault%28%22Auth%22%2C+true%29&type=code

It seems `INGRESS` is a prefix

masayag: added back.

docs/architecture/helm-templates-reference.md
```diff
@@ -92,7 +92,6 @@ This document provides detailed information about the Helm templates used in the
 |----------|-------|-------------|
 | `INGRESS_WEBPORT` | `8080` | Ingress listens on standard port |
 | `INGRESS_METRICSPORT` | `9090` | Prometheus metrics endpoint |
-| `INGRESS_AUTH` | `"false"` | Auth disabled - Gateway handles JWT validation and injects `X-Rh-Identity` |
```
ELK4N4: Revert

ELK4N4 (changes requested):

PR #85, opened 2026-02-08
Gate cloud-only celery workers on cloudProviderSupported, move to clo…

masayag:
…ud-providers/

- Add cost-onprem.koku.cloudProviderSupported helper (hard-coded false) in _helpers-koku.tpl; cloud-only workers deploy only when true (FLPATH-3098)
- Move 9 cloud-only worker deployments into celery/cloud-providers/: refresh, refreshXl, refreshPenalty, download, downloadXl, downloadPenalty, hcs, subsExtraction, subsTransmission
- OCP/supported workers remain in celery/ root

Resource impact: Celery workers before 24 pods / 2.4 cores / 6.5 Gi
→ after (default) 15 pods / 1.5 cores / 3.7 Gi. Saves 9 pods, ~0.9 cores,
~2.8 Gi memory when cloud providers are not enabled.

- Update docs/operations/resource-requirements.md for gated workers and
  default OCP-only footprint.

### DIFF
--- a/docs/operations/resource-requirements.md
+++ b/docs/operations/resource-requirements.md
@@ -8,10 +8,12 @@
 
 | Resource | Minimum (Requests) | Recommended | Maximum (Limits) |
 |----------|-------------------|-------------|------------------|
-| **CPU** | 10.3 cores | 12-14 cores | 17.5 cores |
-| **Memory** | 24 Gi | 32-40 Gi | 52 Gi |
+| **CPU** | 9.4 cores | 11-13 cores | 16 cores |
+| **Memory** | 21 Gi | 28-36 Gi | 48 Gi |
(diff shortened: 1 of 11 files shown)


masayag:
> I recommend to move the workers that relvant to OCP to `ocp/` folder

i think the OCP is always needed. having it in a dedicated folder will create another level that won't necessarily contribute.

docs/operations/resource-requirements.md
```diff
@@ -661,8 +674,8 @@ celery:
 
 ## Version Information
 
-- **Document Version**: 1.2
-- **Based on**: Production deployment observations + Clowder SaaS configuration + OCP-only code analysis (December 2024)
+- **Document Version**: 1.3
```
ELK4N4: Version for a document?

ELK4N4 (changes requested):

ELK4N4 (approved):
/lgtm

PR #99, opened 2026-02-17
[FLPATH-2685] Bring Your Own Infrastructure (BYOI) support for PostgreSQL, Valkey, and Kafka

jordigilh:
## Summary

Adds Bring Your Own Infrastructure (BYOI) support, allowing operators to deploy the chart with externally-managed PostgreSQL, Valkey/Redis, and Kafka instances instead of the bundled ones.

## Rationale

Production deployments often require managed or shared infrastructure (dedicated DBA-managed PostgreSQL, corporate Redis clusters, Strimzi-operated Kafka). The chart previously assumed all infrastructure was bundled, with service names hardcoded to `{release}-database` in init containers, migration jobs, and helper templates. This made it impossible to point the chart at external services without forking templates.

### Helm Templates

The existing `database.server.host` value used an `"internal"` sentinel that the `cost-onprem.database.host` helper resolves to `{release}-database` at render time. For BYOI, operators set `database.deploy: false` and provide a concrete hostname. The same pattern was applied to Valkey (`valkey.deploy`) and Kafka (`kafka.bootstrapServers`).

Key fixes discovered during E2E validation:
- The `waitForDb` init container and both migration jobs (Koku, ROS) had the database hostname **hardcoded** instead of using the `cost-onprem.database.host` helper. These were fixed to use the helper, which resolves correctly in both modes.
- The Koku Kafka helpers had a separate config path (`costManagement.kafka.host/port`) disconnected from the top-level `kafka.bootstrapServers`. These were unified, and the redundant Kafka `ExternalName` service was removed as unnecessary indirection.

### Kafka SASL/TLS Scope Decision

During planning, we discovered that neither Koku nor ROS applications support Kafka SASL/TLS authentication from environment variables in non-Clowder (on-premise) mode. Implementing chart-level SASL/TLS wiring would have no effect without upstream application changes. Kafka BYOI support was therefore scoped to **PLAINTEXT connections only** (matching the current deployment), and the SASL/TLS limitation is documented for future work.

### Install Script

The install script runs **before** any pods exist, so it reads the user's `values.yaml` directly (via `yq`) to determine whether to skip bundled database credential creation when `database.deploy: false`.

### Valkey Persistent Storage

Removed the `valkey.persistence.enabled` toggle -- persistent storage is now always used when deploying the bundled Valkey instance. Ephemeral mode offered no practical benefit and risked Celery task/chord data loss on pod restarts. This was agreed upon during review with @masayag and @ELK4N4.

### Standardize `existingSecret` to `secretName`

Renamed `existingSecret` to `secretName` across all three services (`database.secretName`, `objectStorage.secretName`, `valkey.auth.secretName`). The "existing" qualifier was an implementation detail — in every case the operator provides a secret name and the chart references it. The new name is clearer and consistent across the chart.

Also added a `fail` guard for `valkey.auth`: when `auth.enabled` is true but `secretName` is empty, `helm template`/`helm install` errors out with a clear message instead of rendering a broken manifest.

### Test Strategy: Single Code Path via Pod Introspection

Rather than branching test logic on deployment mode, we leveraged the fact that the Helm templates already resolve `DB_HOST` / `DATABASE_SERVICE_HOST` to a concrete hostname in every app pod, regardless of whether the database is bundled or external. The `database_config` fixture reads this resolved value from a running app pod and discovers the database pod through Kubernetes service endpoint resolution:

1. `printenv DB_HOST` from a ROS or Koku pod (already resolved by Helm)
2. Parse the hostname: simple name (`cost-onprem-database`) stays in chart namespace; FQDN (`postgresql.byoi-infra.svc.cluster.local`) extracts the external namespace
3. `oc get endpoints` resolves the service to the backing pod

This gives a **single code path** that works identically for both modes. In bundled mode, `database_config.namespace` equals `cluster_config.namespace` -- zero behavior change. In BYOI mode, it points to the external namespace.

Only 5 tests that verify **chart-created resources** (database pod exists, database service exists) use a `database_deployed` guard to skip in BYOI mode, since the chart intentionally did not create those resources.

### Helm Templates
- Wrap database templates (StatefulSet, Service, ConfigMap) with `database.deploy` conditional
- Wrap Valkey templates (Deployment, Service, PVC) with `valkey.deploy` conditional
- Fix `waitForDb` init container and migration jobs to use `cost-onprem.database.host` helper
- Fix Koku Valkey host helper for external Valkey with optional auth
- Unify Koku Kafka helpers to derive from top-level `kafka.bootstrapServers`, remove redundant ExternalName service
- Remove `valkey.persistence.enabled` toggle -- persistent storage is always used for bundled Valkey
- Rename `existingSecret` to `secretName` for database, objectStorage, and valkey.auth
- Add `fail` guard when `valkey.auth.enabled` is true but `secretName` is empty

### Install Script
- Add `get_helm_value()` to read user-supplied `values.yaml` pre-deploy via `yq`
- Skip database credentials creation when `database.deploy=false`

### Documentation
- Update `configuration.md` with Kafka SASL/TLS limitation (deferred), hive database notes
- Update `installation.md` with BYOI callouts, examples, and values reference
- Add `byoi-values.yaml` example overlay

### Tests
- Refactor `database_config` fixture with pod introspection discovery (single code path)
- Make `kruize_database_config` reuse discovered DB pod from `database_config`
- Add `database_deployed` fixture for 5 chart-behavior tests
- Update all DB query call sites to use `database_config.namespace`

## Test Plan
- [x] `helm lint` and `helm template` pass with both default and BYOI values
- [x] Full deploy + E2E test in BYOI mode (external PostgreSQL + Valkey in `byoi-infra` namespace)
- [x] 37 passed, 3 skipped (chart-behavior tests correctly skip in BYOI), 0 failed
- [ ] Verify bundled (default) mode is unaffected on CI

## Jira
[FLPATH-2685](https://issues.redhat.com/browse/FLPATH-2685)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -293,20 +293,31 @@ def _get_db_host_from_app_pod(cluster_config: ClusterConfig) -> Optional[str]:
     return None
 
 
-def _parse_k8s_service(hostname: str) -> tuple[Optional[str], Optional[str]]:
-    """Parse a Kubernetes service FQDN into (namespace, service_name).
+@dataclass
+class ParsedService:
+    """A Kubernetes service parsed from its FQDN."""
+
+    namespace: str
+    name: str
+
+
+def _parse_k8s_service(hostname: str) -> Optional[ParsedService]:
+    """Parse a Kubernetes service FQDN into a ParsedService.
 
     Handles forms like:
-      postgresql.byoi-infra.svc.cluster.local  -> ("byoi-infra", "postgresql")
-      postgresql.byoi-infra.svc                -> ("byoi-infra", "postgresql")
-    Returns (None, None) for non-FQDN hostnames.
+      postgresql.byoi-infra.svc.cluster.local  -> ParsedService("byoi-infra", "postgresql")
+      postgresql.byoi-infra.svc                -> ParsedService("byoi-infra", "postgresql")
+    Returns None for non-FQDN hostnames.
     """
     parts = hostname.split(".")
     if len(parts) >= 3 and "svc" in parts:
         svc_idx = parts.index("svc")
         if svc_idx >= 2:
-            return parts[svc_idx - 1], parts[svc_idx - 2]
-    return None, None
+            return ParsedService(
+                namespace=parts[svc_idx - 1],
+                name=parts[svc_idx - 2],
+            )
+    return None
 
 
 def _find_db_pod(namespace: str, service_name: str) -> Optional[str]:
@@ -366,9 +377,11 @@ def database_config(cluster_config: ClusterConfig) -> DatabaseConfig:
     # Step 2: Resolve hostname to namespace + service name
     if ".svc" in db_host:
         # FQDN: "postgresql.byoi-infra.svc.cluster.local"
-        db_namespace, service_name = _parse_k8s_service(db_host)
-        if not db_namespace:
+        parsed = _parse_k8s_service(db_host)
+        if not parsed:
             pytest.skip(f"Cannot parse k8s service from DB host: {db_host}")
+        db_namespace = parsed.namespace
+        service_name = parsed.name
     else:
         # Simple name: "cost-onprem-database" (same namespace)
         db_namespace = cluster_config.namespace


testetson22:
> @testetson22 @masayag should we make the default setup BYOI in the tests? I figure this will probably be the most common deployment and we can test the other approach (current deployment) locally as part of our own testing. Thoughts?

This would essentially bring us back to the old way, from a testing perspective, where we installed all the infra via separate scripts? If so, that seems good if it conforms to the production case.

PR #100, opened 2026-02-19
ci: enforce image parity between helm template and oc-mirror

masayag:
oc-mirror cannot auto-discover images from Helm hooks (pre-install/pre-upgrade), so the PostgreSQL image was missing from disconnected mirrors. Add a CI step that compares images rendered by `helm template` against oc-mirror's mapping.txt and fails with an actionable fix when any image is missing.

- Add additionalImages for postgresql:16 to the CI ImageSetConfiguration
- Add "Verify image parity" step with clear failure diagnostics
- Update disconnected-deployment.md with Required Container Images table, additionalImages in the ImageSetConfiguration example, and callouts explaining why explicit listing is necessary

### DIFF
--- a/.github/workflows/lint-and-validate.yml
+++ b/.github/workflows/lint-and-validate.yml
@@ -5,6 +5,7 @@ on:
     branches: [ main, master ]
     paths:
       - 'cost-onprem/**'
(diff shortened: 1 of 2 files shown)


ELK4N4 (approved):
/lgtm

PR #109, opened 2026-02-25
Allow releases per few commits instead of per PR

masayag:
## Summary
- The version-check workflow no longer fails when the chart version matches the latest release. PRs without a version bump pass with an informational note; only version regressions (lower than latest) fail.
- The release workflow now gates both the GitHub release and OCI push behind a `check-version` job that skips when the release tag already exists, avoiding duplicate runs on pushes that don't bump the chart version.
- When a PR does bump the version, a PR comment announces the upcoming release tag and version.

## Test plan
- [ ] Verify version-check passes on a PR that does not bump the chart version (no_bump result)
- [ ] Verify version-check passes on a PR that bumps the chart version (new_version result, PR comment posted)
- [ ] Verify version-check fails on a PR that lowers the chart version (regression result)
- [ ] Verify release workflow skips when chart version already has a GitHub release
- [ ] Verify release workflow creates GitHub release + OCI push when chart version is new

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,41 @@ on:
       - 'cost-onprem/**'
 
 jobs:
+  check-version:
+    name: Check if version changed
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.chart.outputs.version }}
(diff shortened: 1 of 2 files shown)


masayag:
@jordigilh pls review

PR #141, opened 2026-04-13
FLPATH-3886 Add RC (release candidate) versioning support

masayag:
## Summary

- **bump-version.sh**: Add `--rc` flag with optional `--minor`/`--major` scope qualifiers for creating pre-release versions (e.g., `0.2.20-rc1`). Supports RC increment and promotion to stable via `--patch`.
- **release.yml**: Detect RC versions, mark GitHub releases as pre-release, prevent RC from being marked as latest.
- **version-check.yml**: Detect RC bump type, differentiate PR comments for RC vs stable, add RC suggestion on version regression.
- **release-chart.md**: Document RC workflow lifecycle and promotion steps.

### RC Lifecycle

```
0.2.19 (stable, current)
  → 0.2.20-rc1 (dev/QE)        ./scripts/bump-version.sh --rc
  → 0.2.20-rc2 (dev/QE)        ./scripts/bump-version.sh --rc
  → 0.2.20 (stable, promoted)  ./scripts/bump-version.sh --patch
```

Helm 3 hides pre-release versions from `helm search repo` by default. Users must use `--devel` to see them.

## Test plan

- [x] `--rc` from stable → `0.2.20-rc1`
- [x] `--rc` from RC → `0.2.20-rc2`
- [x] `--rc --minor` from stable → `0.3.0-rc1`
- [x] `--rc --major` from stable → `1.0.0-rc1`
- [x] `--rc --minor` from RC → error (correct)
- [x] `--patch` from RC → `0.2.20` (promote)
- [x] `--minor` from RC → `0.3.0`
- [x] `--major` from RC → `1.0.0`
- [x] `--patch` from stable → `0.2.20` (unchanged behavior)
- [ ] Verify release.yml marks RC as pre-release after merge
- [ ] Verify version-check.yml shows "RC Release Pending" comment on RC PRs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -1,32 +1,12 @@
 #!/bin/bash
 
-# Bump Cost-Onprem Chart Version
-# Bumps the Helm chart version according to semantic versioning.
-# Updates both 'version' and 'appVersion' fields in Chart.yaml.
-#
-# Usage:
-#   ./bump-version.sh --patch
-#   ./bump-version.sh --minor
-#   ./bump-version.sh --major
-#
-# Examples:
-#   # Bump patch: 0.2.9 -> 0.2.10
-#   ./scripts/bump-version.sh --patch
-#
-#   # Bump minor: 0.2.9 -> 0.3.0
-#   ./scripts/bump-version.sh --minor
-#
-#   # Bump major: 0.2.9 -> 1.0.0
-#   ./scripts/bump-version.sh --major
+set -e
 
-set -e  # Exit on any error
-
-# Color codes for output
 RED='\033[0;31m'
 GREEN='\033[0;32m'
 YELLOW='\033[1;33m'
 BLUE='\033[0;34m'
-NC='\033[0m' # No Color
+NC='\033[0m'
 
 echo_info() {
     echo -e "${BLUE}[INFO]${NC} $1"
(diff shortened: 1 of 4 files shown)


scripts/bump-version.sh
```diff
@@ -41,92 +21,156 @@ echo_error() {
 }
 
 usage() {
-    echo "Usage: $0 --patch | --minor | --major"
+    echo "Usage: $0 --patch | --minor | --major | --rc [--patch|--minor|--major]"
```
ELK4N4: I think `patch`/`minor`/`major` should be positional args (no `--` prefix) since they're required. Only `--rc` should be a flag since it's optional. This also follows how tools like `cargo-release` and `npm version` do it:

```bash
bump-version.sh patch              # 1.0.1 -> 1.0.2
bump-version.sh patch --rc         # 1.0.1 -> 1.0.2-rc1
bump-version.sh rc                 # 1.0.2-rc1 -> 1.0.2-rc2
bump-version.sh patch              # 1.0.2-rc2 -> 1.0.2 (promote)
```

ELK4N4 (commented):
Have you considered promoting an RC to stable via a tag instead of a version-bump PR?

Once an RC passes QE, the code is already reviewed — the promote step is purely mechanical. Instead of opening a PR just to strip `-rcN`, you could tag the same commit as stable:

```bash
git tag cost-onprem-0.2.20
git push origin cost-onprem-0.2.20
```

The release workflow would then publish the stable release and auto-commit updating `Chart.yaml` to `0.2.20`.

scripts/bump-version.sh
masayag: Thanks for the suggestion. I considered this but I think the current all-flags approach is the better fit here:

1. **Backward compatibility** — `--patch`/`--minor`/`--major` are already in use and referenced in docs. Switching to positional args is a breaking change unrelated to the RC feature.
2. **Repo consistency** — other scripts in this repo (e.g., `deploy-test-cost-onprem.sh`) use `--` prefixed flags throughout. Positional args would be the odd one out.
3. **Internal consistency** — the proposal mixes positional and flag styles (`bump-version.sh patch --rc`). The current approach (`--rc --minor`) keeps everything uniform.
4. **`npm version`/`cargo-release` analogy** — those are subcommands within larger CLI frameworks, not standalone bash scripts. Different conventions apply.

No functional gain from the change, so I'd prefer to keep the current style.

masayag:
Re: [tag-based promotion suggestion](https://github.com/insights-onprem/cost-onprem-chart/pull/141#pullrequestreview-4099625808)

Good question — I did consider this. The tag-based approach has a few issues that make the one-line PR preferable:

1. **Chart.yaml divergence** — Helm packages embed the version from `Chart.yaml`, not from the git tag. Tagging a commit as `cost-onprem-0.2.20` while `Chart.yaml` still says `0.2.20-rc2` produces a chart with the wrong version baked in. The "auto-commit" fix means CI pushing directly to `main` without review.

2. **Branch protection** — if `main` has required reviews or status checks, a CI-generated commit either bypasses them or needs a special mechanism (PAT, GitHub App token, protection override). That's operational complexity we don't need.

3. **Two release paths** — the release workflow would need a second trigger (tag push) alongside the existing one (push to main). Two paths means more surface area for bugs and more to reason about during incidents.

4. **Audit trail** — a PR for promotion gives a clear record in PR history (who, when, reviewable diff). A tag push is less visible.

The cost of a one-line PR is negligible, and "every release goes through a PR" is a simple invariant that's easy to reason about. I'd prefer to keep it.

ELK4N4 (approved):
/lgtm

PR #142, opened 2026-04-13
Enhanced component update detection, PR staging, and test recommendation

testetson22:
Rework the automated component update workflow to produce **RC-ready PRs** instead of tracking-only PRs.

In previous iterations, the workflow only detected that upstream `latest` tags had changed, recorded them in `component-updates.json`, and created a PR that required a human to manually update `values.yaml` and `Chart.yaml`. Now the workflow does the full prep:

- **Detects upstream updates** by comparing pinned image tags' digests against `latest` digests via the Quay API
- **Patches `values.yaml`** with resolved concrete tags (short commit SHAs, not `latest`)
- **Bumps `Chart.yaml`** RC version (e.g., `0.2.20-rc4` → `0.2.20-rc5`)
- **Creates a release-ready PR** with one automated PR at a time (closes/deletes old ones)
- **Posts test recommendations** with specific Prow `/test` commands and upstream CI verification reminders

### Changed files

| File | What changed |
|------|-------------|
| `scripts/qe/check-components.sh` | Rewritten: extracts all pinned images, Quay API digest comparison, `values.yaml` patching, `Chart.yaml` RC bump, smart tag resolution (prefers commit SHAs over `sha256-` prefixed tags), scaffolded `skopeo` support for Red Hat registries |
| `.github/workflows/check-components.yml` | Single-PR lifecycle (close old `automated/component-update-*` PRs before creating new one), commits chart files, release-ready PR body with upstream CI reminder |
| `scripts/qe/post-test-advisor-comment.sh` | Maps profile to Prow `/test` command (in code block to prevent accidental triggers), adds collapsible upstream CI verification section |
| `scripts/qe/recommend-tests.sh` | New `prow_command` output variable |
| `component-updates.json` | Removed `update_history` (redundant now that chart changes are committed directly) |

## Test plan

- [x] `bash -n` syntax check passes on all 3 scripts
- [x] Workflow YAML validates (`yaml.safe_load`)
- [x] `MODE=list-versions` — extracts all 10 images, Quay API returns digests, Red Hat registry images skipped
- [x] `MODE=detect-updates` — detects real upstream updates, patches `values.yaml` correctly, bumps `Chart.yaml` RC
- [x] Tag resolution prefers short commit SHAs (`df40716`) over `sha256-` prefixed tags
- [x] Missing `values.yaml` produces non-zero exit (validates existing CI test expectation)
- [x] `act` dry-run of `check-components.yml` passes
- [x] `act` full run of `list-versions` mode succeeds end-to-end in container


chadcrum (dismissed):

masayag (commented):
Overall looks good, but cursor with gpt-5.3 had few comments:

Thanks for the substantial automation improvements here — I found a few behavioral issues worth fixing before merge:

1) **High: `has_updates` can remain true even when no image tag patch succeeded**
In `run_detect_updates`, `has_updates=\"true\"` is set **before** `patch_values_tag` succeeds. If patching fails and the loop continues, the script can still bump `Chart.yaml` and proceed toward PR creation without any actual component update applied.

2) **High: unresolved digest->tag lookups are cached as current, which can suppress real updates**
When `quay_resolve_tag` returns empty, the script still writes `latest_digest` to cache. On the next run that digest matches cache, so we stop retrying tag resolution until digest changes again. This can permanently skip a valid update if resolution failed transiently or due to pagination.

3) **Medium: component detection in `recommend-tests.sh` can bind a tag change to the wrong repository**
The values diff parsing uses `grep -B5 <tag_line> ... | tail -1` across the full diff. If multiple added tag lines share the same value, the repository context can be mis-associated, causing incorrect IQE profile recommendations.

Suggested direction:
- Set `has_updates=true` only after successful `patch_values_tag`.
- Do not write digest cache on unresolved tag; retry on next run.
- Parse values diff in hunk-local context (or parse structured YAML before/after) instead of global grep windows.

chadcrum (approved):

PR #146, opened 2026-04-20
[FLPATH-4073] Deploy insights-rbac as RBAC v1 authorization backend

jordigilh:
## Summary

Adds insights-rbac (RBAC v1) as the authorization backend for **both Koku and ROS** in the on-prem Helm chart, replacing Kessel/ReBAC for permission checks. This is the Scenario B prototype tracked in FLPATH-4073.

- Deploys insights-rbac API (Gunicorn) + Celery worker alongside existing services
- Provisions `costonprem_rbac` PostgreSQL database via init script
- Wires Koku to call RBAC via `RBAC_SERVICE_*` env vars in `commonEnv`
- **Wires ROS API to call RBAC** via `RBACHOST`, `RBACPORT`, `RBACPROTOCOL` env vars — RBAC is **always enabled** (hardcoded `RBAC_ENABLE=true`), with no option to disable it
- Sets `ENHANCED_ORG_ADMIN=False` so all users go through RBAC (no admin bypass)
- Migration + seed job runs as Helm hook (weight `-5`, before Koku at `0`)
- NetworkPolicy restricts RBAC ingress to Koku API/Masu pods only
- ServiceMonitor enables Prometheus metrics scraping

### ROS RBAC Integration

RBAC enforcement on the ROS API is **unconditional** — there is no `rbacEnabled` toggle in `values.yaml`. This aligns with the upstream SaaS behavior where ROS always delegates authorization to insights-rbac.

Key changes for ROS:
- `RBAC_ENABLE=true` is hardcoded in the ROS API deployment template (not configurable)
- `RBACHOST`, `RBACPORT`, `RBACPROTOCOL` environment variables are always injected, pointing to the insights-rbac service
- A **`wait-for-rbac` init container** is added to the ROS API pod, ensuring it only starts after the RBAC service is accepting TCP connections — this eliminates the startup race condition between ROS and RBAC
- All ROS and E2E tests are updated to use a password-grant user token (`admin/admin`) instead of the service-account client-credentials token, which RBAC rejects for `type: User` identity headers

## Hardening (validated on OCP cluster)

These changes were added after validating the prototype end-to-end:

| Change | File | Rationale |
|--------|------|-----------|
| Hardcode `is_org_admin: false` | `configmap-envoy.yaml` | **Security fix**: prevents RBAC bypass via user-controlled JWT claim |
| Add `V2_BOOTSTRAP_TENANT` + UUIDs | `_helpers-rbac.tpl` | Prevents `TenantNotBootstrappedError` on RBAC API requests |
| Seed `admin_default` group | `migration-job.yaml` | Admin users get Cost Administrator role via RBAC groups |
| Run `bootstrap_tenants` | `migration-job.yaml` | Creates TenantMapping records required for V1 writes |
| `automountServiceAccountToken: false` | `migration-job.yaml` | Security: no SA token needed in migration pod |

## E2E RBAC Test Suite

New test file (`tests/suites/e2e/test_rbac_access.py`) validates RBAC enforcement with 3 clusters and 3 user personas:
- **alice** (namespace-scoped): sees only "payment" namespace across all clusters
- **bob** (cluster-scoped): sees only cluster-alpha data
- **carol** (full admin): sees everything via Cost Administrator role

10 tests covering permission gates, auto-injection, explicit filters, and unauthenticated access.

## Documentation

- **New**: `docs/operations/rbac-setup.md` — complete RBAC operations guide (architecture, user management, custom policies, cache, troubleshooting)
- **Updated**: `docs/operations/installation.md` — RBAC verification steps, resource table
- **Updated**: `docs/operations/configuration.md` — `rbac.*` values reference, Koku and ROS integration variables

## Architecture

```
User → Keycloak JWT → Envoy Gateway (is_org_admin=false) → Koku API / ROS API
                                                              ↓ (HTTP)
                                                        insights-rbac API ←→ PostgreSQL
                                                              ↓
                                                        Celery Worker ←→ Valkey
```

Both Koku and ROS enforce RBAC permissions by calling insights-rbac with the forwarded identity. The Envoy gateway validates the JWT, injects the `X-Rh-Identity` header, and routes to the appropriate backend.

### ROS API Pod Startup Sequence

```
initContainer: wait-for-database  →  wait-for-kafka  →  wait-for-rbac  →  ros-api container
```

The `wait-for-rbac` init container uses a TCP probe against the RBAC service (`<release>-rbac-api:8000`), matching the same pattern used by all other `waitFor*` init containers in the chart (DB, Kafka, Kruize, Koku).

## Key Design Decisions

| Decision | Rationale |
|----------|-----------|
| `is_org_admin` always `false` | Security: admin access exclusively through RBAC roles, not identity flags |
| `ENHANCED_ORG_ADMIN=False` | Forces all users through RBAC permission checks |
| `BYPASS_BOP_VERIFICATION=True` | MBOP not available on-prem; principals auto-created on first request |
| `admin_default` group per tenant | Grants Cost Administrator to explicit group members |
| `V2_BOOTSTRAP_TENANT=True` | Required for TenantMapping creation (V1/V2 compatibility) |
| `KAFKA_ENABLED=false` | Notifications not needed for authorization flow |
| `CLOWDER_ENABLED=false` | Non-Clowder deployment, explicit env vars for DB/Redis |
| Image pinned to `c2b9ccf` | Reproducible builds; latest active tag as of Apr 13, 2026 |
| Hook weight `-5` | After DB init (`-15`), before Koku migrations (`0`) |
| ROS `RBAC_ENABLE=true` hardcoded | RBAC is always-on for ROS — no toggle. Aligns with SaaS default |
| `wait-for-rbac` init container | Prevents startup race: ROS API blocks until RBAC is ready |
| Password-grant tokens in tests | Service-account tokens rejected by RBAC for `type: User` identities |

## RBAC Permissions Matrix (Operator Reference)

| RBAC Permission | Operations | API Paths |
|----------------|------------|-----------|
| `cost-management:aws.account:read` | GET | `/reports/aws/*`, `/tags/aws/*`, `/forecasts/aws/*` |
| `cost-management:azure.subscription_guid:read` | GET | `/reports/azure/*`, `/tags/azure/*` |
| `cost-management:gcp.account:read` | GET | `/reports/gcp/*`, `/tags/gcp/*` |
| `cost-management:openshift.cluster:read` | GET | `/reports/openshift/*`, `/tags/openshift/*` |
| `cost-management:cost_model:read/write` | GET/POST/PUT/DELETE | `/cost-models/`, `/price-lists/` |
| `cost-management:settings:read/write` | GET/PUT | `/account-settings/*`, `/settings/*` |

> **Note**: ROS API endpoints (`/api/ros/v1/*`) are also protected by these same `cost-management:openshift.*` permissions since ROS uses the same RBAC application and permission set as Koku.

## Test Plan

- [x] `helm lint` passes (verified locally - 0 failures)
- [x] Deploy on OCP cluster with existing Kessel stack
- [x] Verify RBAC migration job completes (DB migrate + seeds)
- [x] Verify RBAC API responds to `GET /api/rbac/v1/status/`
- [x] Create RBAC group + role binding via Django ORM
- [x] Verify Koku API call resolves permissions through RBAC
- [x] Verify ROS API call resolves permissions through RBAC
- [x] Verify OCP cost data query respects RBAC permissions (namespace + cluster filtering)
- [x] Verify user without permissions gets 403
- [x] E2E test suite passes (10/10)
- [x] ROS recommendation tests pass with RBAC enabled
- [x] UI validation: users see only authorized data
- [x] `wait-for-rbac` init container blocks until RBAC is ready

## Known Limitations

| ID | Finding | Severity | Status |
|----|---------|----------|--------|
| COR-4 | Explicit `filter[cluster]=<id>` returns 403 despite user being authorized (Koku format mismatch) | Medium | Documented; auto-injection works correctly |
| OPS-1 | Single-replica RBAC (production should use ≥2 + PDB) | Low | Documented in config guide |
| OPS-4 | No circuit breaker on Koku→RBAC calls (424 on RBAC outage) | Medium | Documented |
| DES-1 | Migration job inline Python (~80 lines) | Low | Follow-up: extract to management command |
| SEC-3 | `PGSSLMODE=disable` for RBAC DB | Medium | Configurable via values.yaml |

Full adversarial review: [PoC Analysis Gist](https://gist.github.com/jordigilh/c81c73ba411637e24a30acd6a743e5fb)

Ref: [FLPATH-4073](https://redhat.atlassian.net/browse/FLPATH-4073)


cost-onprem/values.yaml
```diff
@@ -194,7 +195,6 @@ costManagement:
       CACHED_VIEWS_DISABLED: "False"
       RETAIN_NUM_MONTHS: "3"
       NOTIFICATION_CHECK_TIME: "24"
-      ENHANCED_ORG_ADMIN: "True"
```
masayag, line 197: can we have this set explicitly to "False" instead of removing it?

cost-onprem/values.yaml
```diff
@@ -636,6 +636,64 @@ database:
...
+# -----------------------------------------------------------------------------
+# RBAC v1 Authorization Backend (insights-rbac)
+# -----------------------------------------------------------------------------
+# insights-rbac: RBAC v1 authorization backend for Koku.
+# Koku delegates all permission checks to this service via HTTP.
+# Deploys an API server (Gunicorn) and a Celery worker for async tasks.
+rbac:
+  image:
+    repository: quay.io/cloudservices/rbac
+    tag: "c2b9ccf"
```
masayag, line 655: nice! no need to maintain our own image for it.

cost-onprem/values.yaml
```diff
@@ -636,6 +636,64 @@ database:
...
+    replicas: 1
+    resources:
+      requests:
+        cpu: 100m
+        memory: 512Mi
+      limits:
+        cpu: 500m
+        memory: 2Gi
+
+  env:
```
masayag: any reason for having this in values.yaml?
this can be added directly to the deployment, but if needed to be reused, then can be added o the helper.

cost-onprem/templates/rbac/networkpolicy.yaml
```diff
@@ -0,0 +1,28 @@
...
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/instance: {{ .Release.Name }}
+      app.kubernetes.io/component: rbac-api
+  policyTypes:
+    - Ingress
+  ingress:
+    - from:
+        - podSelector:
+            matchLabels:
```
masayag, line 23: the PR message says `NetworkPolicy restricts RBAC ingress to Koku API/Masu pods only`

what about ROS?

cost-onprem/templates/_helpers-rbac.tpl
```diff
@@ -0,0 +1,123 @@
...
+      name: {{ include "cost-onprem.rbac.database.secretName" . }}
+      key: rbac-password
+- name: DATABASE_HOST
+  value: {{ include "cost-onprem.rbac.database.host" . | quote }}
+- name: DATABASE_PORT
+  value: {{ .Values.database.server.port | default "5432" | quote }}
+- name: REDIS_HOST
+  value: {{ include "cost-onprem.koku.valkey.host" . | quote }}
+- name: REDIS_PORT
+  value: {{ include "cost-onprem.koku.valkey.port" . | quote }}
```
masayag, line 100: is it guaranteed that Valkey is deployed before RBAC? 
i think we might have the same concern about Valkey readiness before koku as well.

It would be better adding same annotation as we do for postgres
```
  annotations:
    # Deploy database before migrations run (pre-install hook)
    # No delete policy - database persists and remains managed by Helm
    "helm.sh/hook": pre-install,pre-upgrade
    "helm.sh/hook-weight": "-10"
```

But to maintain the order of:
1. Database
2. Valkey
3. RBAC
4. Koku / ROS

WDYT?

cost-onprem/values.yaml
```diff
@@ -636,6 +636,64 @@ database:
...
+    PERMISSION_SEEDING_ENABLED: "True"
+    ROLE_SEEDING_ENABLED: "True"
+    GROUP_SEEDING_ENABLED: "True"
+    DJANGO_LOG_LEVEL: "INFO"
+    RBAC_LOG_LEVEL: "INFO"
+    DJANGO_LOG_FORMATTER: "simple"
+    DJANGO_LOG_HANDLERS: "console"
+    ACCESS_CACHE_ENABLED: "True"
+
+  service:
```
masayag: Any reason to expose this to the user? if not, either move to common helper or hard-coded it.

cost-onprem/templates/gateway/configmap-envoy.yaml
```diff
@@ -208,47 +208,33 @@ data:
                       -- =============================================================================
                       -- Build X-Rh-Identity JSON structure for Koku authentication
                       -- =============================================================================
-                      -- This function constructs the X-Rh-Identity header required by Koku.
+                      -- Constructs the X-Rh-Identity header required by Koku and insights-rbac.
```
masayag, line 223: what about ROS?

masayag (changes requested):

cost-onprem/values.yaml
jordigilh: Done — re-added `ENHANCED_ORG_ADMIN: "False"` to `values.yaml` (commit 2314ea47). The helper no longer hardcodes it; `values.yaml` is the single source of truth.

cost-onprem/values.yaml
jordigilh: Good call — moved `API_PATH_PREFIX` to `_helpers-rbac.tpl` as a hardcoded helper (`cost-onprem.rbac.apiPathPrefix`). It's always `/api/rbac` and shouldn't be user-configurable. Same for `service.type` and `service.port` — both now hardcoded in helpers. Commit 2314ea47.

cost-onprem/templates/rbac/networkpolicy.yaml
jordigilh: Good point — added `ros-api` to the RBAC NetworkPolicy ingress rules. Now Koku, Masu, and ROS API pods can all reach the RBAC service. Commit 2314ea47.

cost-onprem/templates/_helpers-rbac.tpl
jordigilh: Agreed — Valkey ordering should be formalized. Currently the RBAC API has readiness probes and will cycle until Redis is up, but that's not ideal. The migration job (pre-install hook, weight -5) only needs PostgreSQL, so it's unaffected. For the API/worker deployments, we can add an init container to wait for Valkey in a follow-up. The current behavior matches how Koku handles the same dependency today (readiness probe retries).

cost-onprem/values.yaml
jordigilh: Done — removed `service.type` and `service.port` from `values.yaml`. Both are now hardcoded: `ClusterIP` in the service template and port `8000` via the `cost-onprem.rbac.service.port` helper. These are internal-only and should not be user-configurable. Commit 2314ea47.

cost-onprem/templates/gateway/configmap-envoy.yaml
jordigilh: Both Koku and ROS route through the same Envoy gateway, so all requests (including ROS recommendations) receive the X-Rh-Identity header from the Lua filter. Added a comment to the Envoy config explaining this. ROS doesn't currently enforce RBAC permissions but gets the header for future use and audit logging. Also added `ros-api` to the RBAC NetworkPolicy so ROS can call RBAC when it does implement permission checks. Commit 2314ea47.

jordigilh: **Update**: ROS now enforces RBAC permissions — this is no longer "future use." As of commits `a1473cab` through `4e4f0b1b`:

- `RBAC_ENABLE=true` is hardcoded in the ROS API deployment (no toggle)
- `RBACHOST`, `RBACPORT`, `RBACPROTOCOL` env vars are unconditionally injected
- A `wait-for-rbac` init container blocks ROS startup until the RBAC service is ready
- The Envoy configmap comment now reads: "Both services enforce RBAC permissions by calling insights-rbac with the forwarded identity"
- All ROS and E2E tests updated to use password-grant user tokens (admin/admin) instead of service-account tokens
- `ros-api` was already added to the RBAC NetworkPolicy ingress rules in commit 2314ea47

cost-onprem/templates/_helpers-rbac.tpl
jordigilh: Addressed in commit `1efa646b`. Added a `waitForValkey` init container (TCP probe, same pattern as all other `waitFor*` helpers) and applied it to all Valkey consumers:

**RBAC** (newly added `initContainers` — had none before):
- `wait-for-database` → `wait-for-valkey` → RBAC API / Worker

**Koku** (added after existing `combineCA`):
- `prepare-ca-bundle` → `wait-for-valkey` → Koku API, MASU, Listener, Celery Beat, and all 10 Celery Workers

The effective startup order is now formalized:
```
Database → Valkey → RBAC → Koku / ROS
```

`helm lint` passes cleanly.

cost-onprem/templates/rbac/deployment-api.yaml
```diff
@@ -0,0 +1,74 @@
...
+        imagePullPolicy: {{ .Values.rbac.image.pullPolicy }}
+        workingDir: /opt/rbac/rbac
+
+        command: ["gunicorn"]
+        args:
+          - "rbac.wsgi"
+          - "--bind=0.0.0.0:{{ include "cost-onprem.rbac.service.port" . }}"
+          - "--workers=2"
+          - "--threads=2"
+          - "--timeout=120"
```
masayag, line 41: Do you think any of this should be configurable?
Or scaling should be done via the deployment's replica?

cost-onprem/values.yaml
masayag: I was referring to the entire `env` element. The values.yaml file shouldn't be a home for this element.

masayag (changes requested):

jordigilh:
**Addressing remaining review comments:**

---

**Re: gunicorn args configurability (deployment-api.yaml line 41)**

These are intentionally hardcoded and not user-configurable. RBAC is a lightweight internal service (only Koku and ROS call it), so 2 workers / 2 threads is adequate for the expected load. Scaling should be done via `rbac.api.replicas` in `values.yaml`, which is already exposed. If we ever need to tune gunicorn params, we can expose them then.

---

**Re: entire `env` element in values.yaml (line 676)**

Done — moved the entire `rbac.env` block out of `values.yaml` and hardcoded all RBAC environment variables directly in the `cost-onprem.rbac.commonEnv` helper template (`_helpers-rbac.tpl`). These are internal configuration values that should not be user-facing:

- `CLOWDER_ENABLED`, `BYPASS_BOP_VERIFICATION`, `KAFKA_ENABLED`, `PGSSLMODE`
- `DJANGO_LOG_LEVEL`, `RBAC_LOG_LEVEL`, `DJANGO_LOG_FORMATTER`, `DJANGO_LOG_HANDLERS`
- `ACCESS_CACHE_ENABLED`

The `range` loop over `.Values.rbac.env` is removed. `helm lint` passes cleanly. Commit `0b8090d8`.

cost-onprem/templates/rbac/deployment-api.yaml
jordigilh: Agreed — these are intentionally hardcoded and not user-configurable. RBAC is a lightweight internal service (only Koku and ROS call it), so 2 workers / 2 threads is adequate. Scaling should be done via `rbac.api.replicas`, which is already exposed in `values.yaml`. If we ever need to tune gunicorn params, we can expose them at that point.

cost-onprem/values.yaml
jordigilh: Done — moved the entire `rbac.env` block out of `values.yaml` and hardcoded all RBAC environment variables directly in the `cost-onprem.rbac.commonEnv` helper (`_helpers-rbac.tpl`). The `range` loop over `.Values.rbac.env` is removed. These are internal values that should not be user-configurable. Commit `0b8090d8`.

jordigilh:
## Comprehensive E2E Pipeline Audit — Fixes Pushed

After repeated one-at-a-time CI failures, performed a full audit of the e2e pipeline covering `run-iqe-tests.sh`, `deploy-rhbk.sh`, `deploy-test-cost-onprem.sh`, `conftest.py`, all Helm templates (RBAC, gateway, routes, NetworkPolicies), and the Prow CI step configuration.

### Commits

**`4d3f9c93` — fix(iqe): add missing DYNACONF_HTTP__OAUTH_CLIENT_SECRET to IQE pod**
Root cause of the persistent 401 Unauthorized errors across all CI runs. The IQE pod was sending password-grant requests to Keycloak for the confidential `cost-management-ui` client without its `client_secret`. Keycloak correctly rejected these as `unauthorized_client`. Added the missing `DYNACONF_HTTP__OAUTH_CLIENT_SECRET` env var sourced from the `iqe-keycloak-credentials` K8s secret.

**`c0338774` — fix(iqe): fix org_id lookup and add UI client secret validation**
- `run-iqe-tests.sh` was querying Keycloak for `username=test` to resolve org_id, but `deploy-rhbk.sh` creates the test user as `admin`. The lookup always silently failed, falling back to the correct default by coincidence. Fixed to query `username=admin`.
- Added explicit warnings when `UI_CLIENT_SECRET` is empty, so missing Keycloak secrets surface immediately instead of manifesting as opaque 401 errors downstream.
- Upgraded UI client secret extraction failure in `deploy-rhbk.sh` from a silent warning to an error with actionable diagnostic message.

**`a4c09557` — refactor(rhbk): replace grep-based JSON parsing with jq**
All 9 instances of `grep -o` / `cut -d'"'` JSON parsing in `deploy-rhbk.sh` replaced with `jq`. The previous patterns assumed JSON field ordering (e.g., `"id"` appearing before `"clientId"` in the same object) and would silently return empty on any field reordering by Keycloak. Added `jq` as a prerequisite check. Also added `&exact=true` to user list queries to avoid ambiguous matches.

**`b483a426` — fix(rhbk): add openid scope to UI client and increase import timeouts**
- The `cost-management-ui` client was missing the `openid` default scope (the operator client had it). This could cause JWT tokens to be missing standard OIDC claims depending on Keycloak version behavior. Added `openid` to both the UI client and the realm-level `defaultDefaultClientScopes`.
- Increased realm import timeout from 120s → 300s and post-import client poll from 60s → 120s to handle slow CI clusters where the previous timeouts caused secret extraction to run before clients were available.
- Fixed the display summary which incorrectly described the UI client as "Public" when it's actually confidential (`publicClient: false`).

### Verified Correct (no changes needed)

- RBAC Helm templates (commonEnv, deployments, migration job, service, NetworkPolicy)
- Gateway Envoy config (JWT validation, Lua X-Rh-Identity injection, `is_org_admin=false`)
- All NetworkPolicies (gateway, RBAC, ingress, ROS, Kruize)
- Route configuration (`/api` path prefix, TLS edge termination)
- IQE pod DYNACONF env vars (all 28 variables verified)
- Test execution ordering (pytest with `_rbac_bootstrap` runs before IQE)
- IQE smoke filter configuration

cost-onprem/templates/rbac/migration-job.yaml
```diff
@@ -0,0 +1,199 @@
...
+                    perm_count += 1
+            print(f"Seeded {perm_count} cost-management permissions")
+
+            roles = [
+                ("Cost Administrator", "Perform any available operation on cost management resources.", True, False, ["*:*"]),
+                ("Cost Price List Administrator", "Perform read and write operations on cost models.", False, False, ["cost_model:*", "settings:*"]),
+                ("Cost Price List Viewer", "Perform read operations on cost models.", False, False, ["cost_model:read", "settings:read"]),
+                ("Cost Cloud Viewer", "Perform read operations on cost reports related to cloud sources.", False, False, ["aws.account:*", "aws.organizational_unit:*", "azure.subscription_guid:*", "gcp.account:*", "gcp.project:*"]),
+                ("Cost OpenShift Viewer", "Perform read operations on cost reports related to OpenShift sources.", False, False, ["openshift.cluster:*"]),
+            ]
```
masayag: Under the assumption I haven't miss anything in this PR, this section must remain in-sync with koku image's version.
We'll need to think how to make sure this in-sync and simple to update.

cost-onprem/templates/rbac/migration-job.yaml
```diff
@@ -0,0 +1,199 @@
...
+                        ).exists()
+                        if has_cm_access:
+                            policy.roles.remove(role)
+                            removed += 1
+            print(f"Removed {removed} role(s) with cost-management or wildcard access from platform_default groups")
+            CLEANUP_DEFAULTS
+            echo "✓ Platform default cleanup complete"
+
+            echo "=== insights-rbac migration job completed ==="
+
```
masayag, line 197: shouldn't there be roles for the rbac system itself?
the openapi spec has the ability to create roles, e.g. https://github.com/RedHatInsights/insights-rbac/blob/a58c88fc539d085dd16891acb5a47848affdb450/docs/source/specs/openapi.json#L1908

docs/operations/rbac-setup.md
```diff
@@ -0,0 +1,678 @@
...
+| `cost-management.openshift.node` | OCP node name |
+| `cost-management.openshift.project` | OCP namespace/project name |
+| `cost-management.aws.account` | AWS account ID |
+| `cost-management.azure.subscription_guid` | Azure subscription GUID |
+| `cost-management.gcp.account` | GCP billing account |
+| `cost-management.gcp.project` | GCP project ID |
+
+### Important Notes
+
+1. **Custom roles must be created via Django ORM** — the RBAC API rejects custom roles for `cost-management` application via REST API
```
masayag: Why is that rejected?

docs/operations/rbac-setup.md
```diff
@@ -0,0 +1,678 @@
...
+- **Key**: Based on user identity + application
+- **Effect**: Permission changes take up to 5 minutes to propagate
+
+### Django Cache (insights-rbac side)
+
+insights-rbac uses Django's cache framework for internal query results.
+
+### Flushing Caches
+
+After making permission changes that need immediate effect:
```
masayag, line 455: we should create jira issue to provide a better experience to than this, wdyt?

docs/operations/rbac-setup.md
```diff
@@ -0,0 +1,678 @@
...
+    E --> F[Add user Principals<br/>to Groups]
+
+    style A fill:#e1f5fe
+    style F fill:#c8e6c9
+```
+
+When a new org is created in Keycloak:
+
+1. The user's first request to a Koku endpoint triggers tenant creation
+2. Run `bootstrap_tenants` to create the TenantMapping:
```
masayag: who's responsible to run this command?

scripts/deploy-rhbk.sh
```diff
@@ -873,13 +891,58 @@ spec:
               userinfo.token.claim: "false"
           # Note: "access" attribute mapper removed - using ENHANCED_ORG_ADMIN mode
           # All authenticated users are org admins with full access within their org
+      # Public client for IQE tests — iqe_jwt's OIDCAuth.from_basic() does not
+      # send client_secret, so password-grant requires a public client.
+      - clientId: cost-management-iqe
```
masayag: shouldn't this be configurable only to be used by CI/explicitly?

cost-onprem/templates/rbac/migration-job.yaml
jordigilh: Agreed — the 23-permission/5-role seed in the migration job is a snapshot of what `insights-rbac` would generate via `manage.py seeds` if `cost-management` were in its `ROLE_LIST`. That seed needs to track changes in the koku image (e.g. if new permissions are added for a feature).

Two options we should evaluate for keeping this in sync:

1. **Automated check in CI** — a step that renders the migration job's permission list and diffs it against the live `insights-rbac` seeds for the matching image tag. Fails if they diverge.
2. **Switch to `manage.py seeds`** — if we add `cost-management` to `ROLE_LIST` env var on the RBAC deployment, the built-in seed command handles this automatically. Trade-off: we lose the explicit audit trail in the template.

For now, the explicit seed in the template matches the current koku image and gives us a clear view of what gets deployed. Will create a JIRA to track this once we align on the approach.

cost-onprem/templates/rbac/migration-job.yaml
jordigilh: Good point. In this commit I've exposed the RBAC admin API externally at `/api/rbac/` through the Envoy gateway (same JWT auth + X-Rh-Identity injection as all other routes). This means the UI team can now target the RBAC management endpoints for the IAM pages.

The `rbac:*` permissions (for managing roles, groups, principals) are seeded by `manage.py seeds` from the insights-rbac image itself — those are separate from the `cost-management:*` permissions we seed in our migration job. So RBAC self-management permissions already exist.

The deeper question is how the **first admin user** gets these permissions without `is_org_admin=true` (which we hardcode to `false`). In SaaS, `is_org_admin` users bypass RBAC checks entirely and can manage everything. On-prem, we don't have that luxury. See my reply on the rbac-setup.md:413 thread for the full analysis and options — that's where we need your input on the bootstrap approach.

docs/operations/rbac-setup.md
jordigilh: Updated the docs to explain the mechanism. `insights-rbac` gates custom role creation behind the `ROLE_CREATE_ALLOW_LIST` environment variable (empty by default, matching SaaS). The API returns `"Custom roles cannot be created for cost-management"` for any application not on this list.

To enable REST-based role creation, we'd set `ROLE_CREATE_ALLOW_LIST=cost-management` on the RBAC deployment. This would allow the UI (or any authenticated caller) to create/update custom roles via the REST API instead of requiring `kubectl exec` + Django ORM.

**However**, there's a related architectural question I'd like your input on:

Currently, `is_org_admin` is hardcoded to `false` in the Envoy Lua filter. In SaaS, org admins have `is_org_admin=true` which bypasses RBAC checks entirely. The `admin_default` group in our migration job is supposed to auto-assign "Cost Administrator" to org admin users, but since no user ever has `is_org_admin=true`, this group has no effect.

The only working admin bootstrap today is our **test fixture** (`conftest.py:_rbac_bootstrap`), which creates a "CI Test Admin" group via `kubectl exec` + ORM and assigns the admin user. For production, there's currently no automated bootstrap path — an admin would need manual ORM commands.

**Options:**
- **A) Enable `ROLE_CREATE_ALLOW_LIST=cost-management`** + keep `is_org_admin=false` + document manual admin bootstrap via REST API after first deploy
- **B) Set `is_org_admin=true` for specific users** via a Helm value that the Lua filter checks against (e.g. a list of admin usernames)
- **C) Create a Helm hook** that bootstraps the admin group/policy via ORM (similar to what `conftest.py` does, but production-grade)

This needs your guidance since it impacts the production admin experience. What's your preference?

docs/operations/rbac-setup.md
jordigilh: Agreed — added a note in the docs flagging this as a known UX gap with a follow-up improvement planned. Will create a JIRA for a dedicated cache flush mechanism (script or Helm hook) after this round of comments. For most use cases, the TTL-based expiry (300s on Koku side) is sufficient without manual intervention.

docs/operations/rbac-setup.md
jordigilh: Updated the docs to clarify: `bootstrap_tenants --all` runs **automatically** on every `helm install` and `helm upgrade` via the RBAC migration Job (pre-install/pre-upgrade hook, weight -5). The manual command is only needed for **new orgs created between Helm releases** — when a user's first API request creates a tenant that didn't exist at the last upgrade. That's a **cluster administrator** responsibility.

scripts/deploy-rhbk.sh
jordigilh: Good call — removed the IQE client entirely from `deploy-rhbk.sh` (realm import). The IQE public client is now created lazily by `run-iqe-tests.sh` via the Keycloak admin API (`ensure_iqe_keycloak_client()`) only when IQE tests actually run. This keeps test-only infrastructure out of the base Keycloak deployment.

cost-onprem/templates/rbac/migration-job.yaml
jordigilh: JIRA created: [FLPATH-4107](https://redhat.atlassian.net/browse/FLPATH-4107) — Automate RBAC permission/role seed sync with koku image (linked to FLPATH-4073).

docs/operations/rbac-setup.md
jordigilh: JIRA created: [FLPATH-4108](https://redhat.atlassian.net/browse/FLPATH-4108) — Improve RBAC cache flush UX with dedicated script or automation (linked to FLPATH-4073).

docs/operations/rbac-setup.md
masayag: The system should be operational and fully functional after the installation, therefore the selected option(s) should fulfill it.

Option B will enable us to get there. It can be combined with option A (ROLE_CREATE_ALLOW_LIST=cost-management) to also unlock custom role creation via REST.

And once there is an admin configured, they can leverage the UI to administer other users.
I'll let claude reply with its own answer that elaborates a bit more on this.

masayag: ## Analysis of Options

### Problem Decomposition

The question conflates two distinct problems:

1. **Admin bootstrap** (chicken-and-egg): After first deploy, no user has admin access, and you need admin access to grant admin access.
2. **Custom role creation**: The REST API rejects custom roles for `cost-management` unless `ROLE_CREATE_ALLOW_LIST` is set.

Option A only addresses (2). The bootstrap problem (1) is the critical blocker for production.

---

### Ranking

#### 1st: Option B — `is_org_admin=true` for designated users via Helm values

This is the strongest option for on-prem:

- **Declarative and auditable**: Admin usernames live in `values.yaml`, version-controlled, visible in cluster config. The deployer decides who the admins are at install time — which is exactly how on-prem products work.
- **Aligns with upstream design**: The `admin_default` group + auto-assignment mechanism already exists in insights-rbac, specifically designed for `is_org_admin=true` users. Option B makes it work as intended rather than building a parallel path.
- **Minimal code change**: Modify the Envoy Lua filter to read admin usernames from a ConfigMap (populated from Helm values) and set `is_org_admin=true` for matches.
- **Unlocks the REST API path**: Once designated admins have `is_org_admin=true`, they can manage groups, policies, and principals via the RBAC REST API — no `kubectl exec` needed for ongoing operations.

The security concern (username spoofing) is a non-issue in practice: the deployer controls both Keycloak and the Helm values. A username in the admin list is only dangerous if an attacker can create that user in Keycloak, which requires Keycloak admin access — at which point they already own the system.

**Suggestion**: Combine with Option A (`ROLE_CREATE_ALLOW_LIST=cost-management`) to also unlock custom role creation via REST. Together, **B+A** gives designated admins a complete management experience through the API.

#### 2nd: Option C — Helm hook for admin bootstrap

Acceptable but introduces unnecessary coupling and operational overhead:

- **Fragile**: Tight coupling to RBAC's Django ORM schema. If upstream insights-rbac changes model fields, migrations, or the tenant structure, the Job breaks silently or produces corrupt state.
- **Idempotency complexity**: Must handle re-runs on `helm upgrade` correctly — check if the group/policy/principal already exist, handle partial failures, deal with schema version mismatches between the Job's code and the deployed RBAC version.
- **Operational surface area**: Another Job to monitor, debug, and triage in production. If it fails, the admin has no access and must fall back to `kubectl exec` anyway.
- **Doesn't unlock ongoing management**: Even after bootstrap, subsequent user management still requires `kubectl exec` + ORM unless combined with B or A.

The one advantage over B is that it doesn't introduce an `is_org_admin` bypass in the request path — but that's a theoretical purity concern, not a practical one.

#### 3rd: Option A — `ROLE_CREATE_ALLOW_LIST` only

**Insufficient on its own.** It solves custom role creation via REST but doesn't address the bootstrap problem. The RBAC management API endpoints (groups, policies, principals) also require either `is_org_admin=true` or existing RBAC permissions. Without bootstrap, the first admin still needs `kubectl exec` + ORM.

Useful as a complement to B, not as a standalone solution.

---

### Alternative: Option D — `manage.py` management command

If there's resistance to putting `is_org_admin=true` in the request path at all:

- Ship a `manage.py bootstrap_admin --username <user> --org-id <org>` command with the RBAC image
- Documented one-time post-deploy step: `kubectl exec ... -- python manage.py bootstrap_admin --username admin --org-id 12345`
- Simpler than Option C (no Helm hook, no Job, no idempotency concerns)
- Less fragile (uses Django's own ORM within the RBAC codebase, not external)
- Still requires `kubectl exec`, but only once — then the bootstrapped admin can use the REST API (if combined with A) for all subsequent management

This is a pragmatic middle ground if Option B's `is_org_admin` bypass is considered unacceptable, but that concern doesn't hold up under scrutiny for on-prem deployments where the deployer controls both Keycloak and the Helm values.

---

### Recommendation

**B + A** is the recommended path. It's the simplest, most declarative, least fragile, and aligns with both upstream RBAC design and on-prem deployment conventions. The Lua filter change is small and testable. The `admin_default` group mechanism works as upstream intended. And the admin experience post-deploy is entirely REST-based — no `kubectl exec` required for day-2 operations.

cost-onprem/templates/_helpers-rbac.tpl
```diff
@@ -0,0 +1,153 @@
...
+  value: "00000000-0000-4000-a000-000000000003"
+- name: SYSTEM_ADMIN_TENANT_ROLE_UUID
+  value: "00000000-0000-4000-a000-000000000004"
+- name: CLOWDER_ENABLED
+  value: "false"
+- name: BYPASS_BOP_VERIFICATION
+  value: "True"
+- name: KAFKA_ENABLED
+  value: "false"
+- name: PGSSLMODE
```
masayag, line 122: ### SEC-1: `PGSSLMODE=disable` hardcoded, not configurable (High)

The PR description and `docs/operations/configuration.md` claim this is "configurable via values.yaml" (`Set to "require" for production TLS`), but it's hardcoded here with no `{{ .Values }}` reference. An operator cannot enable TLS for the RBAC→PostgreSQL connection without modifying the template.

The other DB clients (Koku, ROS, Kruize) support PGSSLMODE configuration. RBAC should be consistent.

**Recommendation:** Read from values:
```yaml
- name: PGSSLMODE
  value: {{ .Values.rbac.pgsslmode | default "disable" | quote }}
```

cost-onprem/templates/rbac/migration-job.yaml
```diff
@@ -0,0 +1,199 @@
...
+                        name='Cost Admin Default Policy', tenant=tenant, group=grp
+                    )
+                    policy.roles.add(cost_admin_role)
+                    count += 1
+                print(f"Created/updated admin_default group for {count} tenant(s)")
+            ADMIN_DEFAULT_SCRIPT
+            echo "✓ Admin default group seeding complete"
+
+            echo "Bootstrapping tenant mappings for all existing tenants..."
+            python manage.py bootstrap_tenants --all -v 2 || true
```
masayag: ### SEC-2: `bootstrap_tenants` failure silently swallowed (High)

```bash
python manage.py bootstrap_tenants --all -v 2 || true
```

If tenant bootstrap fails, the migration job reports success. RBAC then returns 400 on `/access/` queries (as documented in the troubleshooting section of this very PR), but the operator sees a successful `helm install`. This is a silent failure that surfaces as mysterious 400s in production.

**Recommendation:** Remove `|| true`. If bootstrap fails, the migration should fail so the operator knows to investigate. If there are expected benign failures (e.g. no tenants exist yet on fresh install), handle those specific exit codes explicitly rather than swallowing all errors.

cost-onprem/templates/rbac/migration-job.yaml
```diff
@@ -0,0 +1,199 @@
...
+
+            removed = 0
+            for group in Group.objects.filter(platform_default=True):
+                for policy in Policy.objects.filter(group=group):
+                    for role in policy.roles.all():
+                        has_cm_access = Access.objects.filter(
+                            role=role
+                        ).filter(
+                            Q(permission__application='cost-management')
+                            | Q(permission__resource_type='*')
```
masayag: ### SEC-3: Platform default cleanup query is too broad (Medium)

The wildcard check:
```python
Q(permission__application='cost-management')
| Q(permission__resource_type='*')
```

The second clause matches **any** role with a wildcard `resource_type`, not just cost-management wildcards. If insights-rbac seeds platform_default roles with wildcard permissions for other applications (e.g., `notifications:*:*`), those would also be stripped from platform_default groups.

The same overly broad query appears in `tests/conftest.py` (`_rbac_bootstrap` cleanup).

**Recommendation:** Scope the wildcard check to cost-management:
```python
Q(permission__application='cost-management')
| Q(permission__application='cost-management', permission__resource_type='*')
```

docs/operations/installation.md
```diff
@@ -739,6 +741,49 @@ oc rsh -n cost-onprem deployment/cost-onprem-ingress -- \
...
+kubectl get pods -l app.kubernetes.io/component=rbac-api -n cost-onprem
+
+# Verify seeded roles exist
+RBAC_POD=$(kubectl get pod -l app.kubernetes.io/component=rbac-api -n cost-onprem -o jsonpath='{.items[0].metadata.name}')
+kubectl exec -it $RBAC_POD -n cost-onprem -- \
+  python manage.py shell -c "from management.models import Role; roles = Role.objects.filter(name__startswith='Cost'); print(f'Cost roles: {[r.name for r in roles]}')"
+
+# Test RBAC API status endpoint
+kubectl exec -it $RBAC_POD -n cost-onprem -- \
+  curl -s http://localhost:8080/api/rbac/v1/status/ | python3 -m json.tool
```
masayag: ### COR-1: Documentation references port 8080, but templates use 8000 (High)

The RBAC service port is hardcoded to **8000** in `cost-onprem.rbac.service.port` helper, and gunicorn binds to `0.0.0.0:8000`. But the documentation references **8080** in multiple places:

- `docs/operations/installation.md:760,765` — `curl -s http://localhost:8080/...`
- `docs/operations/configuration.md:601,637` — `port: 8080`, `RBAC_SERVICE_PORT` default `8080`
- `docs/operations/rbac-setup.md` — `http://localhost:8080/...`

Operators following the docs would get connection refused. All `8080` references in RBAC context should be `8000`.

docs/operations/configuration.md
```diff
@@ -583,6 +583,93 @@ ui:
         memory: "64Mi"
 ```
 
+### RBAC Configuration
```
masayag, line 587: ### COR-2: Configuration docs describe values.yaml fields that don't exist (Medium)

The RBAC configuration example shows several fields that don't match the actual `values.yaml`:

| Documented field | Actual state |
|---|---|
| `rbac.enabled: true` | Field doesn't exist; RBAC has no toggle |
| `rbac.api.replicaCount` | Actual field is `rbac.api.replicas` |
| `rbac.api.port: 8080` | Port is hardcoded to 8000 in helper, not configurable |
| `rbac.env` block | Was moved out of `values.yaml` per earlier review feedback; now hardcoded in `_helpers-rbac.tpl` |
| `PGSSLMODE: "disable"` with comment `Set to "require" for production TLS` | Not actually configurable (see SEC-1) |

This documentation would mislead operators into thinking they can configure things that are actually hardcoded.

cost-onprem/templates/rbac/migration-job.yaml
```diff
@@ -0,0 +1,199 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "cost-onprem.labels" . | nindent 4 }}
+    app.kubernetes.io/component: rbac-migration
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  backoffLimit: 0
```
masayag: ### OPS-1: `backoffLimit: 0` — no retry on transient failures (Medium)

If the migration job fails for any reason beyond the DB readiness check (OOM, Django migration conflict, seed script error, network flake during `bootstrap_tenants`), there's no retry. Combined with `helm.sh/hook-delete-policy: before-hook-creation`, a failed migration blocks the helm install/upgrade and requires manual intervention.

**Recommendation:** `backoffLimit: 1` or `2` to handle transient failures. The job is already idempotent (uses `get_or_create` throughout), so retries are safe.

cost-onprem/templates/rbac/networkpolicy.yaml
```diff
@@ -0,0 +1,32 @@
...
+    {{- include "cost-onprem.labels" . | nindent 4 }}
+    app.kubernetes.io/component: networkpolicy
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/instance: {{ .Release.Name }}
+      app.kubernetes.io/component: rbac-api
+  policyTypes:
+    - Ingress
+  ingress:
```
masayag, line 16: ### OPS-2: NetworkPolicy may block Envoy→RBAC traffic (Medium)

The review discussion ([comment](https://github.com/insights-onprem/cost-onprem-chart/pull/146#discussion_r3144587013)) states:
> "I've exposed the RBAC admin API externally at `/api/rbac/` through the Envoy gateway"

However, this NetworkPolicy only allows ingress from `cost-management-api`, `cost-management-masu`, and `ros-api`. The Envoy gateway pod is **not** in the allow list.

If the RBAC REST API is intended to be reachable through the gateway (for admin role management), this NetworkPolicy would block that traffic. Either:
1. Add the gateway pod selector to the ingress rules, or
2. Clarify that the Envoy→RBAC route is not yet implemented in this PR and the review comment was forward-looking

cost-onprem/templates/rbac/deployment-worker.yaml
```diff
@@ -0,0 +1,50 @@
...
+        image: "{{ include "cost-onprem.rbac.image" . }}"
+        imagePullPolicy: {{ .Values.rbac.image.pullPolicy }}
+        workingDir: /opt/rbac/rbac
+
+        command: ["celery"]
+        args:
+          - "-A"
+          - "rbac.celery"
+          - "worker"
+          - "--pool=solo"
```
masayag, line 40: ### OPS-3: Celery worker `--pool=solo` with 4:1 memory overcommit (Medium)

`--pool=solo` means single-threaded, sequential task processing. For cache invalidation and async RBAC operations, this could become a bottleneck under load.

Separately, the worker has a 4:1 memory overcommit ratio (512Mi request / 2Gi limit). On memory-constrained nodes, the scheduler places the pod based on the 512Mi request, but the pod can burst to 2Gi — risking OOM kills when the node is under pressure.

tests/suites/sources/test_sources_api.py
```diff
@@ -483,22 +484,30 @@ def test_non_admin_source_creation_returns_424(
             },
         )
 
-        assert response.status_code == 424, f"Expected 424, got {response.status_code}: {response.text[:200]}"
+        assert response.status_code in (201, 403, 424), (
```
masayag, line 488: ### DES-1: Test assertions weakened to accept contradictory outcomes (Medium)

This test and `test_missing_email_in_identity_rejected` below had their assertions significantly weakened:

- Here: Was `assert status == 424`, now `assert status in (201, 403, 424)` — the test name says creation should be restricted, but status 201 (created successfully) is accepted.
- `test_missing_email_in_identity_rejected` (line 507): Was `assert status == 401`, now `assert status in (401, 403, 500)` — accepting 500 (server error) as valid for a "rejected" test.

Accepting 3+ status codes per test means these can't catch regressions where behavior changes. A test that passes for both "allowed" and "denied" outcomes provides no confidence.

**Recommendation:** Use conditional assertions based on whether RBAC is deployed, or split into separate tests for each expected behavior.

tests/conftest.py
```diff
@@ -653,6 +710,342 @@ def org_id(cluster_config: ClusterConfig, keycloak_config: KeycloakConfig) -> st
         return "org1234567"
 
 
+# =============================================================================
+# RBAC Bootstrap (ensures CI service account has permissions)
+# =============================================================================
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _rbac_bootstrap(cluster_config: ClusterConfig, keycloak_config: KeycloakConfig, gateway_url: str):
```
masayag, line 761: ### DES-2: `_rbac_bootstrap` fixture is ~340 lines of inline ORM scripts (Medium)

This session-scoped autouse fixture contains:
- 3 separate `exec_in_pod_raw` calls executing inline Python scripts
- A 100+ line diagnostic script querying both RBAC API and Koku directly
- Multiple cache flush operations
- Error paths that `logger.warning` and continue silently

If any inline script fails silently (which several error paths allow), tests may pass for wrong reasons or fail with misleading errors. The diagnostic section alone is ~80 lines of Python embedded in an f-string with escaped braces.

**Recommendation:** Extract the bootstrap, cleanup, and diagnostic scripts into standalone `.py` files (or management commands) that can be tested and maintained independently. The fixture would then just call those scripts.

cost-onprem/templates/rbac/migration-job.yaml
```diff
@@ -0,0 +1,199 @@
...
+
+            echo "Running Django migrations..."
+            python manage.py migrate --noinput
+            echo "✓ Migrations complete"
+
+            echo "Seeding permissions, roles, and groups..."
+            python manage.py seeds --skip-notifications
+            echo "✓ Built-in seeding complete"
+
+            echo "Seeding cost-management permissions and roles (from RedHatInsights/rbac-config)..."
```
masayag, line 65: ### DES-3: Seed data drift risk between migration job and RBAC image (Medium)

The 23-permission / 5-role seed here is a point-in-time snapshot that must stay synchronized with:
1. The RBAC image version (`c2b9ccf`)
2. Koku's DRF permission checks
3. ROS's permission checks

If any of these three change independently, the system breaks silently — a new permission required by Koku won't exist in RBAC, or a renamed permission in RBAC won't match Koku's expectations.

There's no automated validation (CI check, helm test, or startup probe) that the seeded permissions match what the applications expect.

**Recommendation:** Add a helm test or CI validation that queries the RBAC `/access/` endpoint and verifies all expected permissions exist. This catches drift before it reaches production.

masayag (changes requested):
## Adversarial Due Diligence Review

Comprehensive review covering security, correctness, operational robustness, design quality, and maintainability.

**Scope**: 48 files, ~3000 additions across RBAC infrastructure, security model, test infrastructure, E2E tests, documentation, and Keycloak configuration.

**Positives**: The RBAC test matrix (3 users × 3 clusters × multiple filter scenarios) is thorough. The `is_org_admin=false` hardcoding is the correct security decision. Init container ordering is well-thought-out. The documentation is extensive (even if it has accuracy issues noted below).

### Findings Summary

| ID | Severity | Category | File | Finding |
|----|----------|----------|------|---------|
| SEC-1 | High | Security | `_helpers-rbac.tpl` | `PGSSLMODE=disable` hardcoded, not configurable |
| SEC-2 | High | Security | `migration-job.yaml` | `bootstrap_tenants` failure silently swallowed |
| SEC-3 | Medium | Security | `migration-job.yaml` | Platform default cleanup query too broad |
| COR-1 | High | Correctness | `installation.md` | Docs reference port 8080, templates use 8000 |
| COR-2 | Medium | Correctness | `configuration.md` | Docs describe values.yaml fields that don't exist |
| OPS-1 | Medium | Operational | `migration-job.yaml` | `backoffLimit: 0` — no retry on transient failures |
| OPS-2 | Medium | Operational | `networkpolicy.yaml` | NetworkPolicy may block Envoy→RBAC if admin API exposed |
| OPS-3 | Medium | Operational | `deployment-worker.yaml` | Celery `--pool=solo` with 4:1 memory overcommit |
| DES-1 | Medium | Design | `test_sources_api.py` | Test assertions weakened to accept contradictory outcomes |
| DES-2 | Medium | Design | `conftest.py` | `_rbac_bootstrap` is ~340 lines of inline ORM scripts |
| DES-3 | Medium | Design | `migration-job.yaml` | Seed data drift risk between migration job and RBAC image |

See inline comments for details and recommendations.

cost-onprem/templates/rbac/migration-job.yaml
masayag: This seems to be covered by the next comment in which we decided on creating a Jira issue to address this by a CI job or other method.

cost-onprem/templates/rbac/networkpolicy.yaml
jordigilh: Good catch — this is already handled. The RBAC NetworkPolicy explicitly allows ingress from the `gateway` component:

```yaml
matchLabels:
  app.kubernetes.io/instance: {{ .Release.Name }}
  app.kubernetes.io/component: gateway
```

So Envoy (which carries `app.kubernetes.io/component: gateway`) is already in the allow list. No change needed.

cost-onprem/templates/rbac/deployment-worker.yaml
jordigilh: `--pool=solo` was chosen because the RBAC Celery worker handles a small number of async tasks (cache invalidation, role seeding notifications) that don't benefit from prefork parallelism. Solo avoids the memory overhead of child processes, which is appropriate for a single-replica deployment.

The Koku workers under `cost-management/celery/` use the default prefork pool because they handle high-throughput data processing queues.

For the memory ratio: 512Mi request / 2Gi limit matches the ceiling used by heavy Koku workers (summary, priority). The burst headroom accommodates occasional large Django querysets during role synchronization without requiring a permanently high request that wastes scheduler capacity.

Happy to tighten the limit if you have a specific target in mind — we haven't observed OOMKills with the current settings.

scripts/deploy-rhbk.sh
jordigilh: This is already handled by existing code structure:

- The IQE client is **not** part of the RHBK realm import (`deploy-rhbk.sh`) — it only ships the operator and UI clients.
- The IQE client is created on-demand by `scripts/run-iqe-tests.sh` via the `ensure_iqe_keycloak_client` function (line 337+), which uses the Keycloak admin API to create a public client at test time.

So in a production deployment (no IQE tests), the IQE client never exists. It only appears when an operator explicitly runs the IQE test suite.

cost-onprem/templates/_helpers-rbac.tpl
jordigilh: Addressed in commit `a1773b82`. SEC-1: Wired PGSSLMODE to `database.server.sslMode` — no longer hardcoded.

cost-onprem/templates/rbac/migration-job.yaml
jordigilh: Addressed in commit `a1773b82`. SEC-2: Replaced `|| true` with `set +e` / capture `$?` / `set -e` — bootstrap failures are now visible while platform_default cleanup is guaranteed to run.

cost-onprem/templates/rbac/migration-job.yaml
jordigilh: Addressed in commit `a1773b82`. SEC-3: Simplified Q clause to `permission__application='cost-management'` only — removed the over-broad `resource_type='*'` filter.

docs/operations/installation.md
jordigilh: Addressed in commit `a1773b82`. COR-1: Fixed all 5 port references from 8080 to 8000 across installation.md, rbac-setup.md, and configuration.md.

docs/operations/configuration.md
jordigilh: Addressed in commit `a1773b82`. COR-2: Aligned the RBAC section with actual `values.yaml` — fixed image repo, renamed `replicaCount` → `replicas`, removed phantom `enabled`/`port`/`env` fields, added note about commonEnv helper.

cost-onprem/templates/rbac/migration-job.yaml
jordigilh: Addressed in commit `a1773b82`. OPS-1: Changed `backoffLimit` from 0 to 1.

tests/suites/sources/test_sources_api.py
jordigilh: Addressed in commit `a1773b82`. DES-1: Tightened assertion from `(401, 403, 500)` back to `(401, 403)` — a 500 is a bug, not expected behavior.

tests/conftest.py
jordigilh: Addressed in commit `a1773b82`. DES-2: Extracted all 3 inline scripts to `tests/rbac_bootstrap_scripts.py` with `repr()`-hardened interpolation. Fixture now calls `render_bootstrap_script()`, `CLEANUP_SCRIPT`, and `render_diag_script()`.

docs/operations/rbac-setup.md
jordigilh: Addressed in commit `a1773b82`. Implemented Option B+A:

**Option B** (`is_org_admin`): Added `jwtAuth.orgAdminUsernames` in `values.yaml`. The Envoy Lua filter now renders a Helm-templated lookup table — usernames in the list get `is_org_admin:true` in the `X-Rh-Identity` header, which triggers insights-rbac's `admin_default` mechanism (granting Cost Administrator via the seeded "Cost Admin Default" group). Empty list (default) keeps `is_org_admin:false` for everyone.

**Option A** (`ROLE_CREATE_ALLOW_LIST`): Added `rbac.roleCreateAllowList` in `values.yaml`, conditionally injected as an env var on the RBAC API deployment when non-empty.

jordigilh:
@masayag Identified a gap in the RBAC user bootstrap flow: `deploy-rhbk.sh` creates the `admin/admin` Keycloak user, but RBAC permissions are only granted by the test fixture — meaning a user following the install guide would log in successfully but get 403 on every API call.

Proposed fix: a standalone `sync-rbac-admin.sh` post-install script + doc updates.

Full analysis and plan: https://gist.github.com/jordigilh/a8f2dd943621ff71142832916fc792ab

Let me know if you have any concerns before I implement.

tests/rbac_bootstrap_scripts.py
```diff
@@ -0,0 +1,232 @@
...
+from django.db.models import Q
+from django.core.cache import cache
+
+group_removed = 0
+for group in Group.objects.filter(platform_default=True):
+    for policy in Policy.objects.filter(group=group):
+        for role in policy.roles.all():
+            has_cm = Access.objects.filter(role=role).filter(
+                Q(permission__application='cost-management')
+                | Q(permission__resource_type='*')
```
masayag: **SEC-3 residual (minor):** The migration job query was correctly narrowed to `permission__application='cost-management'` only, but the `CLEANUP_SCRIPT` here still has the broad wildcard clause:
```python
Q(permission__application='cost-management')
| Q(permission__resource_type='*')
```

This is test-only code and the practical impact is near-zero (no other RBAC applications exist in this on-prem deployment), but it's inconsistent with the production fix. Consider aligning for consistency.

tests/suites/sources/test_sources_api.py
```diff
@@ -483,22 +484,29 @@ def test_non_admin_source_creation_returns_424(
...
```
masayag, line 488: **DES-1 residual (minor):** `test_missing_email_in_identity_rejected` was correctly tightened from `(401, 403, 500)` to `(401, 403)`, but this test still accepts `(201, 403, 424)`. Accepting both 201 (allowed) and 403 (denied) means the test can't distinguish between RBAC granting and denying access — it passes regardless of authorization outcome.

Non-blocking, but worth a follow-up to narrow this based on whether the non-admin identity has any RBAC group membership.

masayag (approved):
## Re-Review: All High Issues Resolved, Two Minor Items Remain

Re-reviewed the 4 new commits (`a1773b82`..`adcc9265`) addressing the 11 findings from the previous review.

### Resolution Tracker

| ID | Finding | Status | Notes |
|----|---------|--------|-------|
| SEC-1 | PGSSLMODE hardcoded | **Resolved** | Now tracks `database.server.sslMode` |
| SEC-2 | bootstrap_tenants swallowed | **Resolved** | `set +e` / capture `$?` / `set -e` — failures visible in logs |
| SEC-3 | Broad wildcard query | **Partially resolved** | Migration job fixed; test script still has broad query (see below) |
| COR-1 | Port 8080 vs 8000 | **Resolved** | All references fixed |
| COR-2 | Docs vs values.yaml drift | **Resolved** | Aligned with actual fields, added commonEnv note |
| OPS-1 | backoffLimit: 0 | **Resolved** | Changed to 1 |
| OPS-2 | NetworkPolicy gap | **Resolved** | Gateway pod selector added |
| OPS-3 | Celery --pool=solo | **Acknowledged** | Appropriate for low-volume async tasks |
| DES-1 | Weakened test assertions | **Partially resolved** | Email test tightened; source creation test unchanged (see below) |
| DES-2 | Inline ORM scripts | **Resolved** | Extracted to `rbac_bootstrap_scripts.py` |
| DES-3 | Seed data drift | **Acknowledged** | Follow-up Jira issue created |

### New Code Review (4 commits)

The new additions are well-structured:

- **admin-bootstrap-job.yaml**: Clean post-install hook, gated by `rbac.bootstrapAdmin.enabled` (default false), idempotent, `backoffLimit: 1`. Implements the Option B+A recommendation from the earlier discussion.
- **Envoy Lua org admin lookup**: O(1) hash table rendered from `jwtAuth.orgAdminUsernames`, values properly escaped via Helm `quote`. Clean implementation.
- **RBAC API Envoy route**: `/api/rbac/` routed to `rbac-api-backend` with same JWT validation as other routes. NetworkPolicy updated to match.
- **`roleCreateAllowList`** in values.yaml: Enables REST-based custom role creation (Option A).
- **sync-rbac-admin.sh**: Clean manual fallback with proper arg parsing and `set -euo pipefail`.
- **IQE client moved to `run-iqe-tests.sh`**: Correct separation — test client created on-demand, not shipped in realm import.

### Remaining Items (non-blocking)

See inline comments for the two minor outstanding items.

cost-onprem/templates/rbac/migration-job.yaml
lcouzens: I may have missed this, but are we planning to add roles for sources too? In the Saas today there is sources roles specifically since you may want to grant a user (Or more likely a service account) the ability to create sources (for new clusters) but not give them full org admin. I think as of right now only an org admin could create the cluster sources correct?

scripts/sync-rbac-admin.sh
```diff
@@ -0,0 +1,109 @@
...
+
+echo "Creating Tenant, Principal, Group, and Policy..."
+kubectl exec -n "${NAMESPACE}" "${RBAC_POD}" -- \
+  python /opt/rbac/rbac/manage.py shell -c "
+from api.models import Tenant
+from management.models import Group, Policy, Role, Principal
+from django.core.cache import cache
+
+username = '${USERNAME}'
+org_id = '${ORG_ID}'
```
masayag: **Minor inconsistency**: The fallback script uses single-quoted Python strings for interpolated values:
```python
username = '${USERNAME}'
```

while the hook (`admin-bootstrap-job.yaml`) uses double-quoted strings:
```python
username = "${BOOTSTRAP_USERNAME}"
```

Single-quoted strings break on usernames containing `'` (e.g. `o'brien`). The hook's double-quoted approach is more robust. Not a practical concern for typical Keycloak usernames, but worth aligning for consistency since these two scripts do the same thing.

scripts/sync-rbac-admin.sh
```diff
@@ -0,0 +1,109 @@
...
+
+cache.clear()
+print(f'User \"{username}\" granted Cost Administrator for org={org_id}')
+"
+
+echo ""
+echo "Running bootstrap_tenants for TenantMapping/V2 records..."
+kubectl exec -n "${NAMESPACE}" "${RBAC_POD}" -- \
+  python /opt/rbac/rbac/manage.py bootstrap_tenants --org-id "${ORG_ID}" --force || \
+  echo "WARNING: bootstrap_tenants returned non-zero (non-fatal)"
```
masayag: **Minor inconsistency**: Error handling pattern differs from the hook.

Hook uses `set +e` / capture `$?` / `set -e`:
```bash
set +e
python manage.py bootstrap_tenants --org-id ... --force
bootstrap_rc=$?
set -e
if [ $bootstrap_rc -ne 0 ]; then
  echo "WARNING: ... exited with code $bootstrap_rc ..."
fi
```

This script uses `|| echo`:
```bash
python ... bootstrap_tenants ... || echo "WARNING: ..."
```

Both are functionally correct (non-fatal), but the hook preserves the exit code in logs while `||` masks it. Consider aligning with the hook pattern for debuggability.

masayag (dismissed):
## Review: Last 3 Commits (Admin Bootstrap Implementation)

Reviewed commits `32a41109`..`adcc9265` against the agreed plan from the [RBAC Admin User Sync gist](https://gist.github.com/jordigilh/a8f2dd943621ff71142832916fc792ab).

### Plan Compliance

All 15 requirements from the agreed Approach C are implemented correctly:

- Hook annotations (`post-install,post-upgrade`, weight 5, `before-hook-creation`) ✓
- Conditional on `rbac.bootstrapAdmin.enabled` (default `false`) ✓
- Same RBAC image + `commonEnv` pattern as migration job ✓
- DB TCP wait loop (300s timeout) ✓
- 9-step ORM logic in correct order (public tenant → Cost Admin role lookup → Tenant → Group → Policy → Principal → cache clear) ✓
- `bootstrap_tenants --org-id --force` with non-fatal error handling ✓
- `backoffLimit: 1` ✓
- `values.yaml` defaults match `deploy-rhbk.sh` admin user ✓
- `sync-rbac-admin.sh` fallback with arg parsing ✓
- `deploy-rhbk.sh` summary updated ✓
- `installation.md` documents both Option A (Helm) and Option B (script) ✓
- `rbac-setup.md` `is_org_admin` inconsistency fixed (2 occurrences) ✓
- Test fixture untouched ✓

### Two Minor Inconsistencies (non-blocking)

See inline comments.

tests/rbac_bootstrap_scripts.py
jordigilh: Done — removed `| Q(permission__resource_type='*')` from `CLEANUP_SCRIPT`. Now matches the migration job's narrowed query. Commit `8f3c9b75`.

tests/suites/sources/test_sources_api.py
jordigilh: Done — tightened `test_non_admin_source_creation` from `(201, 403, 424)` to `(403, 424)`. A non-admin user without `sources:*:*` should not get 201.

Note: koku's sources endpoint currently uses `AllowAny` so this test will fail until enforcement is added. Filed [FLPATH-4132](https://redhat.atlassian.net/browse/FLPATH-4132) for that follow-up. Commit `8f3c9b75`.

cost-onprem/templates/rbac/migration-job.yaml
jordigilh: Good catch — added in commit `8f3c9b75`:

- `sources:*:*` permission seeded in the migration job
- "Sources administrator" role with `admin_default=true` (matching [SaaS rbac-config](https://github.com/RedHatInsights/rbac-config/blob/master/configs/prod/roles/sources.json))

The admin_default group logic now uses `Role.objects.filter(admin_default=True)` so org admins automatically get both Cost Administrator and Sources administrator.

However, koku's built-in `/sources` endpoint (`SourcesViewSet`) currently uses `permission_classes = (AllowAny,)` — the role is seeded but not enforced until a koku-side code change lands. Filed [FLPATH-4132](https://redhat.atlassian.net/browse/FLPATH-4132) for that.

scripts/sync-rbac-admin.sh
jordigilh: Done — switched to double-quoted Python strings (`"$\{USERNAME}"`) matching the hook. Also refactored the script to use `admin_default_roles` query (same pattern as the hook) so it grants all `admin_default` roles including the newly-added Sources administrator. Commit `8f3c9b75`.

scripts/sync-rbac-admin.sh
jordigilh: Done — replaced `|| echo "WARNING..."` with `set +e` / capture `$?` / `set -e`, consistent with the hook. Commit `8f3c9b75`.

masayag:
pls rebase so we can merge this.

masayag (approved):

PR #153, opened 2026-04-26
docs: disconnected guide — discover images with helm/oc-mirror

masayag:
## Summary

Replace the static **Required Container Images** table in `docs/operations/disconnected-deployment.md` with repeatable commands: `helm template` + awk for chart images, and `oc-mirror --v2` dry-run plus `mapping.txt` for the full mirror plan. Tags drift with releases, so generated lists stay accurate.

## Other changes

- Point readers at `.github/workflows/lint-and-validate.yml` `additionalImages` for hook/parity alignment.
- Adjust CI parity failure messages to reference the doc instead of asking to duplicate the removed table.

### DIFF
--- a/docs/examples/disconnected/imageset-config-cost-onprem-with-prerequisites.sample.yaml
+++ b/docs/examples/disconnected/imageset-config-cost-onprem-with-prerequisites.sample.yaml
@@ -1,49 +1,64 @@
-# SAMPLE ImageSetConfiguration: cost-onprem Helm chart plus OLM content for
-# dependencies installed by this repository's scripts:
-#   - AMQ Streams: ./scripts/deploy-kafka.sh (Subscription name amq-streams)
-#   - RHBK:        ./scripts/deploy-rhbk.sh (Subscription name rhbk-operator)
-# Optional OpenShift Data Foundation (ODF) packages are commented below; use
-# them only if you deploy ODF for S3-compatible storage on the cluster.
+# COMPREHENSIVE ImageSetConfiguration: OpenShift 4.20 platform + cost-onprem + all prerequisites
 #
-# Before mirroring:
-# 1. This sample targets OpenShift 4.20 (redhat-operator-index:v4.20). Change
-#    the catalog tag if your cluster is a different minor version.
-# 2. Confirm package channels exist on your version (OperatorHub on a
-#    connected cluster, or Red Hat documentation for disconnected mirroring).
-# 3. Uncomment the odf-operator entry only when mirroring ODF; verify the
-#    channel name in OperatorHub for your release (example uses stable-4.20).
+# This configuration mirrors everything needed for a complete disconnected deployment:
+#   - OpenShift 4.20.12 platform images
+#   - cost-onprem Helm chart and container images
+#   - OLM operators for all deployment scripts:
+#     * AMQ Streams (deploy-kafka.sh)
+#     * RHBK (deploy-rhbk.sh)
+#     * Local Storage + ODF operators (for S3-compatible storage)
+#   - Additional images for hook-only containers and script dependencies
+#
+# Tested working configuration for OpenShift 4.20.
 #
 # Usage:
 #   oc-mirror --v2 -c imageset-config-cost-onprem-with-prerequisites.sample.yaml file://mirror-output
 #
-# Chart version: edit mirror.helm.repositories[].charts[].version as in
-# imageset-config-cost-onprem.yaml.
+# Before mirroring:
+# 1. Edit mirror.helm.repositories[].charts[].version to match your target chart release
+# 2. Adjust platform.channels[].minVersion/maxVersion for your specific OCP patch version
+# 3. Verify operator channels exist for your OpenShift version in OperatorHub
+#
 apiVersion: mirror.openshift.io/v2alpha1
 kind: ImageSetConfiguration
+# OCP 4.20: platform + redhat-operator-index + Marketplace index images + cost-onprem + RHODF-related operators + AMQ Streams.
 mirror:
+  platform:
+    architectures:
+      - amd64
+    graph: true
+    channels:
+      - name: stable-4.20
+        type: ocp
+        minVersion: 4.20.12
+        maxVersion: 4.20.12
   helm:
     repositories:
       - name: cost-onprem
         url: https://insights-onprem.github.io/cost-onprem-chart
         charts:
           - name: cost-onprem
-            version: "0.2.20-rc3"
+            version: "0.2.20-rc4"
+  additionalImages:
+    # PostgreSQL image used by both cost-onprem (Helm hook) and deploy-rhbk.sh (Keycloak DB):
+    - name: registry.redhat.io/rhel10/postgresql-16:10.1
+    # AWS CLI for S3 bucket creation (install-helm-chart.sh):
+    - name: docker.io/amazon/aws-cli:latest
   operators:
     - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.20
       packages:
         - name: amq-streams
           channels:
+            - name: stable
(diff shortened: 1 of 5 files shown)


docs/operations/disconnected-deployment.md
```diff
@@ -21,46 +22,92 @@ In disconnected environments, clusters have no direct internet access. The `oc-m
...
+flags so defaults match offline discovery.
+
+```bash
+cd /path/to/cost-onprem-chart/cost-onprem
+
+helm template cost-onprem . > /tmp/cost-onprem-rendered.yaml
+
+awk -F': ' '/^[[:space:]]+image:/{gsub(/"/,"",$2); print $2}' \
+  /tmp/cost-onprem-rendered.yaml | sort -u
+```
```
ELK4N4: Should we move this snippet to a separate bash file in case it will be changed in the future?
Also, consider using `yq` for yaml parsing

docs/operations/disconnected-deployment.md
```diff
@@ -21,46 +22,92 @@ In disconnected environments, clusters have no direct internet access. The `oc-m
...
+cat > /tmp/imageset-config.yaml <<'EOF'
+apiVersion: mirror.openshift.io/v2alpha1
+kind: ImageSetConfiguration
+mirror:
+  helm:
+    local:
+      - name: cost-onprem
+        path: /path/to/cost-onprem-chart/cost-onprem
+  additionalImages:
+    # Copy from .github/workflows/lint-and-validate.yml for hook-only images.
```
ELK4N4: We can extract `ImageSetConfiguration` resource to a separate yaml file, so the user and the CI can use the same file

docs/operations/disconnected-deployment.md
masayag: this was copied from the workflow file - I don't think we need both options (oc-mirror and this method). I'll remove this and let's stick to oc-mirror only.

docs/operations/disconnected-deployment.md
masayag: this is a good recommendation and makes a lot of sense.

docs/examples/disconnected/imageset-config-cost-onprem.sample.yaml
```diff
@@ -0,0 +1,31 @@
...
+# install from https://insights-onprem.github.io/cost-onprem-chart (see
+# cost-onprem/Chart.yaml in this repository for the current chart version).
+#
+# For local chart mirroring instead of the Helm repo, replace mirror.helm with:
+#   helm:
+#     local:
+#       - name: cost-onprem
+#         path: /path/to/cost-onprem-chart/cost-onprem
+#
+# Keep additionalImages aligned with .github/workflows/lint-and-validate.yml.
```
ELK4N4, line 17: `.github/workflows/lint-and-validate.yml` should use this file

docs/examples/disconnected/imageset-config-cost-onprem-with-prerequisites.sample.yaml
```diff
@@ -0,0 +1,49 @@
+# SAMPLE ImageSetConfiguration: cost-onprem Helm chart plus OLM content for
+# dependencies installed by this repository's scripts:
```
ELK4N4: Do we need to support these dependencies?
To my understanding, this is the customer's responsibility.

ELK4N4 (changes requested):
`docs/examples/disconnected/imageset-config-cost-onprem-with-prerequisites.sample.yaml`
`docs/examples/disconnected/imageset-config-cost-onprem.yaml`

The first one is with `.sample.yaml` suffix. It's not consistent

docs/examples/disconnected/imageset-config-cost-onprem-with-prerequisites.sample.yaml
masayag: I agree these are prerequisites of the chart, but in order to install everything in air-gapped, we need to provide a recommendation on how to do that to make that experience more pleasant.

E.g. if this project should be demoed in air-gapped env, it will be useful to include it here.

docs/examples/disconnected/imageset-config-cost-onprem.sample.yaml
masayag: we can't use it, since the workflow relies on local chart repo, while this file targets a specific version from the helm registry.

see the one from the workflow:
```
        kind: ImageSetConfiguration
        mirror:
          helm:
            local:
              - name: cost-onprem
                path: ./cost-onprem
          additionalImages:
            - name: quay.io/insights-onprem/postgresql:16
```

masayag:
@y-first can you have a look and try this in a disconnected env?

ELK4N4 (dismissed):

docs/operations/disconnected-deployment.md
```diff
@@ -125,9 +150,9 @@ This configures the cluster to pull images from the mirror registry instead of t
 Install the chart from the mirrored registry. Use the install script with the local chart:
 
 ```bash
-# Option A: Use the mirrored chart directly
+# Option A: Use the mirrored chart directly (--version must match mirrored chart)
 helm install cost-onprem oci://mirror.example.com:5000/cost-onprem/cost-onprem \
```
y-first: @masayag Are we sure this is correct? oc-mirror does not create OCI images for Helm, it just mirrors the chart dependencies. When testing these instructions as written (with the correct path to the mirror registry) this Helm install fails. Unless there is a missing step here.

masayag: hmm...
oc-mirror isn't responsible for creating the chart OCI image. This is done by [our CI](https://github.com/insights-onprem/cost-onprem-chart/blob/main/.github/workflows/release.yml#L113) for now (in the future we'll do that using konflux).
However, If we provide oci image as input to oc-mirror, I'd expect it to mirror that image as well.
I don't see that option happens, so perhaps we'll need to include the chart image under the additionalmages.

y-first: The source of the oci image, and/or the instructions for mirroring it, should be clearer in the documentation. The implication as written is that the oc-mirror will create it.

masayag: Agree, I'll clarify it

masayag:
Before we merge this PR, let's get https://github.com/insights-onprem/cost-onprem-chart/pull/154 in and then I'll update this one.
We are using two different postgresql image versions (one for cost and another for RHBK), while we can use the same version and simplify the images to mirror and reduce the need to use own build of postgresq.

cc @y-first this will affect the ImageSetConfiguration file.

docs/examples/disconnected/imageset-config-cost-onprem-with-prerequisites.sample.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - name: cost-onprem
+        url: https://insights-onprem.github.io/cost-onprem-chart
+        charts:
+          - name: cost-onprem
+            version: "0.2.20-rc4"
+  additionalImages:
+    # PostgreSQL image used by both cost-onprem (Helm hook) and deploy-rhbk.sh (Keycloak DB):
+    - name: registry.redhat.io/rhel10/postgresql-16:10.1
+    # AWS CLI for S3 bucket creation (install-helm-chart.sh):
+    - name: docker.io/amazon/aws-cli:latest
```
y-first, line 46: mismatch here between the other sample without pre-reqs that uses `amazon/aws-cli:latest` without docker.io. I needed to add the docker.io in my setup but that may not be the case for everyone

masayag: thanks for highlighting this.
I added it to the other file to be explicit.

jordigilh (approved):
lgtm

PR #167, opened 2026-05-11
fix: derive is_org_admin from Keycloak org-admin realm role

jordigilh:
## Summary

- **Single source of truth**: replaces the static `orgAdminUsernames` list with Keycloak realm roles — the Envoy Lua filter now reads `realm_access.roles` from the JWT to set `is_org_admin` in `X-Rh-Identity`, making Keycloak the authoritative source for admin status
- **User provisioning**: `values.yaml` gains a structured `realmUsers` list (admin + viewer) with `orgAdmin` flag; `deploy-rhbk.sh` provisions these users into Keycloak with the `org-admin` realm role and adds `roles` to both clients' `defaultClientScopes` for forward-compatibility
- **Three-layer test coverage**: JWT claim tests (admin has `org-admin` in `realm_access.roles`, viewer does not), gateway identity propagation tests (admin gets `cost-management:*:*` from RBAC, viewer does not), and Playwright tests (admin sees Settings content, viewer sees "access denied")
- **Security fix**: sanitized `$VALUES_FILE` path in `deploy-rhbk.sh` to prevent shell injection

## Test plan

- [ ] Verify `deploy-rhbk.sh -f cost-onprem/values.yaml` provisions both admin (with org-admin role) and viewer (without) in Keycloak
- [ ] Verify admin/admin can access Settings page after deployment
- [ ] Verify viewer/viewer sees "You do not have access to Settings in cost management"
- [ ] Run `pytest -m auth` — JWT realm role and gateway identity tests pass
- [ ] Run `pytest -m ui` — Playwright Settings access tests pass
- [ ] Verify no regression on existing e2e/auth/sources test suites

### DIFF
--- a/cost-onprem/templates/_helpers-rbac.tpl
+++ b/cost-onprem/templates/_helpers-rbac.tpl
@@ -115,6 +115,9 @@ Provides DB, Redis, and application configuration for non-Clowder mode.
   value: "00000000-0000-4000-a000-000000000004"
 - name: CLOWDER_ENABLED
   value: "false"
+  {{/* On-prem has no Back Office Proxy (BOP/mbop); bypass its identity
+       verification so RBAC accepts the X-Rh-Identity header directly from
+       Envoy. Required for on-prem; SaaS uses BOP for this check. */}}
 - name: BYPASS_BOP_VERIFICATION
   value: "True"
 - name: KAFKA_ENABLED
--- a/cost-onprem/templates/_helpers-rbac.tpl
+++ b/cost-onprem/templates/_helpers-rbac.tpl
@@ -148,6 +148,51 @@ RBAC service port (always 8000 — internal only, not user-configurable).
 {{- end -}}
 
 {{/*
+Resolve the bootstrap admin identity from jwtAuth.realmUsers.
+Finds the first entry with orgAdmin=true and returns username, orgId,
+accountNumber. Falls back to rbac.bootstrapAdmin values if realmUsers
+is empty or has no orgAdmin entry.
+*/}}
+{{- define "cost-onprem.rbac.bootstrapAdmin.username" -}}
+{{- $found := false -}}
+{{- range .Values.jwtAuth.realmUsers -}}
+  {{- if and (not $found) .orgAdmin -}}
+    {{- $found = true -}}
+    {{- .username -}}
+  {{- end -}}
+{{- end -}}
+{{- if not $found -}}
+  {{- .Values.rbac.bootstrapAdmin.username | default "admin" -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "cost-onprem.rbac.bootstrapAdmin.orgId" -}}
+{{- $found := false -}}
+{{- range .Values.jwtAuth.realmUsers -}}
+  {{- if and (not $found) .orgAdmin -}}
+    {{- $found = true -}}
+    {{- .orgId -}}
+  {{- end -}}
+{{- end -}}
+{{- if not $found -}}
+  {{- .Values.rbac.bootstrapAdmin.orgId | default "org1234567" -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "cost-onprem.rbac.bootstrapAdmin.accountNumber" -}}
+{{- $found := false -}}
+{{- range .Values.jwtAuth.realmUsers -}}
+  {{- if and (not $found) .orgAdmin -}}
+    {{- $found = true -}}
+    {{- .accountNumber -}}
+  {{- end -}}
+{{- end -}}
+{{- if not $found -}}
+  {{- .Values.rbac.bootstrapAdmin.accountNumber | default "7890123" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 RBAC database secret name. Reuses the main DB credentials secret
 with RBAC-specific keys, or a dedicated secret.
 */}}
--- a/cost-onprem/templates/_helpers-rbac.tpl
+++ b/cost-onprem/templates/_helpers-rbac.tpl
@@ -153,19 +153,22 @@ RBAC service port (always 8000 — internal only, not user-configurable).
 {{/*
 Resolve the bootstrap admin identity from jwtAuth.realmUsers.
 Finds the first entry with orgAdmin=true and returns username, orgId,
-accountNumber. Falls back to rbac.bootstrapAdmin values if realmUsers
-is empty or has no orgAdmin entry.
+accountNumber. When bootstrapAdmin.enabled is true and no orgAdmin
+user exists, the template fails with an actionable error.
 */}}
 {{- define "cost-onprem.rbac.bootstrapAdmin.username" -}}
 {{- $found := false -}}
 {{- range .Values.jwtAuth.realmUsers -}}
   {{- if and (not $found) .orgAdmin -}}
     {{- $found = true -}}
-    {{- .username -}}
+    {{- .username | required "jwtAuth.realmUsers: orgAdmin entry must have a username" -}}
   {{- end -}}
 {{- end -}}
 {{- if not $found -}}
-  {{- .Values.rbac.bootstrapAdmin.username | default "admin" -}}
+  {{- if .Values.rbac.bootstrapAdmin.enabled -}}
+    {{- fail "rbac.bootstrapAdmin.enabled is true but no orgAdmin:true entry found in jwtAuth.realmUsers" -}}
+  {{- end -}}
+  {{- "admin" -}}
 {{- end -}}
 {{- end -}}
 
@@ -174,11 +177,11 @@ is empty or has no orgAdmin entry.
 {{- range .Values.jwtAuth.realmUsers -}}
   {{- if and (not $found) .orgAdmin -}}
     {{- $found = true -}}
-    {{- .orgId -}}
+    {{- .orgId | required "jwtAuth.realmUsers: orgAdmin entry must have an orgId" -}}
   {{- end -}}
 {{- end -}}
 {{- if not $found -}}
-  {{- .Values.rbac.bootstrapAdmin.orgId | default "org1234567" -}}
+  {{- "org1234567" -}}
 {{- end -}}
 {{- end -}}
 
@@ -187,11 +190,11 @@ is empty or has no orgAdmin entry.
 {{- range .Values.jwtAuth.realmUsers -}}
   {{- if and (not $found) .orgAdmin -}}
     {{- $found = true -}}
-    {{- .accountNumber -}}
+    {{- .accountNumber | required "jwtAuth.realmUsers: orgAdmin entry must have an accountNumber" -}}
   {{- end -}}
 {{- end -}}
 {{- if not $found -}}
-  {{- .Values.rbac.bootstrapAdmin.accountNumber | default "7890123" -}}
+  {{- "7890123" -}}
 {{- end -}}
 {{- end -}}
 
--- a/cost-onprem/templates/gateway/configmap-envoy.yaml
+++ b/cost-onprem/templates/gateway/configmap-envoy.yaml
@@ -375,8 +375,10 @@ data:
                           -- Use replace to avoid duplicate headers
                           request_handle:headers():replace("X-Rh-Identity", xrhid_b64)
 
-                          -- Forward the original Authorization header as X-Bearer-Token
-                          -- Try both capitalization variants for case-insensitive lookup
+                          -- Forward the raw JWT as X-Bearer-Token for backends that
+                          -- need it (e.g., oauth-proxy session refresh). Envoy has
+                          -- already validated the signature via jwt_authn; backends
+                          -- MUST NOT trust this header for authz — use X-Rh-Identity.
                           local auth_header = request_handle:headers():get("authorization")
                           if not auth_header then
                             auth_header = request_handle:headers():get("Authorization")
--- a/cost-onprem/templates/gateway/configmap-envoy.yaml
+++ b/cost-onprem/templates/gateway/configmap-envoy.yaml
@@ -227,11 +227,10 @@ data:
                       -- permissions from the "Cost Admin Default" group (seeded by the migration
                       -- job with "Cost Administrator" role = cost-management:*:*).
                       --
-                      -- In on-prem there is no BOP. The org_admin_usernames table below is
-                      -- rendered from jwtAuth.orgAdminUsernames in values.yaml. Only listed
-                      -- usernames get is_org_admin=true. When the list is empty (default),
-                      -- is_org_admin is false for everyone and admin access must be granted
-                      -- through explicit RBAC group membership.
+                      -- In on-prem, is_org_admin is derived from the Keycloak "org-admin"
+                      -- realm role. If the JWT contains "org-admin" in realm_access.roles,
+                      -- the user gets is_org_admin=true. Keycloak is the single source of
+                      -- truth — assign/remove the role in the Keycloak admin console.
                       --
                       -- Both Koku and ROS route through this gateway and receive the same
                       -- X-Rh-Identity header. Both services enforce RBAC permissions by
@@ -241,12 +240,17 @@ data:
                       -- middleware returns 403 without it and there is no bypass.
                       -- =============================================================================
 
-                      -- Org admin lookup table (rendered from jwtAuth.orgAdminUsernames)
-                      local org_admin_usernames = {
-                        {{- range .Values.jwtAuth.orgAdminUsernames }}
-                        [{{ . | quote }}] = true,
-                        {{- end }}
-                      }
+                      -- Check if JWT realm_access.roles contains a specific role
+                      local function has_realm_role(jwt_data, role_name)
+                        local realm_access = jwt_data["realm_access"]
+                        if realm_access == nil then return false end
+                        local roles = realm_access["roles"]
+                        if roles == nil then return false end
+                        for _, role in ipairs(roles) do
+                          if role == role_name then return true end
+                        end
+                        return false
+                      end
 
                       local function build_xrhid_json(org_id, account_number, user_type, username, email, is_org_admin)
                         local function escape_json(str)
@@ -363,7 +367,7 @@ data:
                             email = username .. "@example.com"
                           end
 
-                          local is_org_admin = org_admin_usernames[username] or false
+                          local is_org_admin = has_realm_role(jwt_data, "org-admin")
                           local xrhid_json = build_xrhid_json(org_id, account_number, "User", username, email, is_org_admin)
                           local xrhid_b64 = request_handle:base64Escape(xrhid_json)
 
--- a/cost-onprem/templates/rbac/admin-bootstrap-job.yaml
+++ b/cost-onprem/templates/rbac/admin-bootstrap-job.yaml
@@ -41,9 +41,9 @@ spec:
             MAX_WAIT=300
             ELAPSED=0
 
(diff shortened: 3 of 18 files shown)


masayag (commented):
## PR Review: derive is_org_admin from Keycloak org-admin realm role

### Architecture

Moving from a static Helm-rendered Lua lookup table to a runtime JWT `realm_access.roles` check is the correct design. Decouples admin status management from `helm upgrade` cycles and makes Keycloak the single source of truth for Day 2 operations.

Test strategy (JWT claims → gateway identity propagation → Playwright UI) is solid.

---

### Flag & Script Wiring Issues

**1. `deploy-test-cost-onprem.sh` never passes `-f` to `deploy-rhbk.sh` — viewer user not provisioned in CI (High)**

`deploy-test-cost-onprem.sh:436`:
```bash
if ! execute_script "${SCRIPT_DEPLOY_RHBK}"; then
```

No `-f` argument. So `deploy-rhbk.sh` runs with `VALUES_FILE=""`, hits the default path in `get_realm_users_json()`, and provisions only the single hardcoded admin user. The viewer user defined in `values.yaml` `realmUsers` is never created.

New tests depend on the viewer user:
- `test_non_admin_jwt_lacks_org_admin_realm_role` — uses `_obtain_token(viewer/viewer)`, which calls `pytest.skip` on non-200 → test silently skips in CI, providing zero coverage
- `test_non_admin_lacks_cost_management_access` — uses `obtain_user_jwt_token_for(viewer/viewer)`, which calls `pytest.fail` on non-200 → **test fails in CI**
- `test_non_admin_settings_page_shows_access_denied` — Playwright `_login_context(viewer/viewer)` tries to log in to a non-existent Keycloak user → login page hangs, **test times out in CI**

Either `deploy-test-cost-onprem.sh` needs to pass `-f` with the values file, or the default fallback in `get_realm_users_json()` needs to include the viewer user to match what `values.yaml` declares.

---

**2. `display_summary()` not updated — stale output post-provisioning (Low)**

`deploy-rhbk.sh:1474-1489` (unchanged by PR):
```bash
echo_info "  User: admin"
echo_info "    Password: admin"
echo_info "      org_id: org1234567 ..."
echo_info "    The defaults (username=admin, orgId=org1234567, accountNumber=7890123)"
echo_info "    match this user."
```

When `-f` is used with custom `realmUsers`, the summary still hardcodes the admin user info. After provisioning users from a custom values file, the summary is inaccurate.

---

**3. `-f` flag silently ignored if placed after command (Low)**

`deploy-rhbk.sh` uses a two-stage parser: `while` loop consumes `-f`, then `case` dispatches the command. The `*` case breaks out of the while loop:

- `deploy-rhbk.sh -f values.yaml` → works ✓
- `deploy-rhbk.sh -f values.yaml validate` → works ✓
- `deploy-rhbk.sh validate -f values.yaml` → `while` breaks on `validate`, `-f values.yaml` are never parsed. Silent ignore.

The help text documents the correct order (`Usage: $0 [-f <values.yaml>] [command]`), but this is inconsistent with `sync-rbac-admin.sh` where `-f` can appear anywhere in the arg list.

---

### Script Bugs

**4. `create_realm_users` ignores per-user failure, always reports success (Medium)**

`create_realm_users()` ignores `create_or_update_user` return codes:

```bash
while [ $i -lt "$user_count" ]; do
    ...
    create_or_update_user "$KEYCLOAK_URL" "$ACCESS_TOKEN" "$user_obj"  # RC ignored
    i=$((i + 1))
done
echo_success "✓ Realm user provisioning complete"  # Always printed
```

A user provisioning failure (empty `USER_ID`, role assignment failure) goes unreported. The script exits 0 and the operator sees green output with a partially configured Keycloak.

Fix: accumulate a failure counter and exit non-zero / print a summary.

---

**5. IQE identity discovery hardcoded to `username=admin` (Medium)**

Both `run-iqe-tests.sh:325` and `run-iqe-tests-local.sh:484`:

```bash
USER_JSON=$(curl -sk ".../users?username=admin&exact=true" ...)
```

The PR claims `realmUsers` is the canonical source for admin identity, but these scripts don't read from it to determine which username to query. If someone names their admin user differently in `realmUsers`, these scripts silently fall back to default `org1234567` / `7890123`.

Fix: either accept `-f values.yaml` to discover the admin username, or document that the admin username must be `admin` for IQE to work.

---

**6. `sync-rbac-admin.sh -f` silently falls back to defaults on parse failure (Medium)**

```bash
admin_json=$(python3 -c "..." "$VALUES_FILE" 2>/dev/null) && {
    ...
} || echo "WARNING: Could not parse admin from $VALUES_FILE, using defaults"
```

When `-f` is explicitly provided but parsing fails or no `orgAdmin: true` entry exists, the script warns and proceeds with `admin` / `org1234567` / `7890123`. This script mutates RBAC state (creates Tenant, Principal, Group, Policy) — falling back silently when the user explicitly provided a file is wrong.

Note: `deploy-rhbk.sh` got hardened for the PyYAML-missing case (patch 4) but `sync-rbac-admin.sh` didn't get the same treatment.

Fix: when `-f` is provided, any parse/selection failure should be a hard error.

---

**7. `2>/dev/null` swallows PyYAML error in `get_realm_users_json` (Low)**

```bash
users_json=$(python3 -c "..." "$VALUES_FILE" 2>/dev/null)
```

The Python code explicitly prints `"ERROR: PyYAML is required..."` to stderr, but `2>/dev/null` suppresses it. The `echo_error` added in patch 4 partially compensates, but the original diagnostic is lost.

Fix: remove `2>/dev/null` or capture stderr for diagnostics.

---

### Code Quality

**8. Nested function definition (`_obtain_admin_token` inside `create_realm_users`) (Low)**

Bash functions are global regardless of lexical nesting. This works incidentally but is misleading — it looks scoped but isn't. The inner function references `$KEYCLOAK_URL`, `$ADMIN_USERNAME`, `$ADMIN_PASSWORD` from the outer function's locals, which works only because bash doesn't have closures (it reads them at call time from the surrounding scope). Fragile if `create_realm_users` is ever called in a different context.

Recommendation: move `_obtain_admin_token` to top level, pass URL/credentials as arguments.

---

**9. Default passwords in committed `values.yaml` (Low)**

`password: admin` and `password: viewer` ship as defaults. The WARNING comment is there — low severity given pre-GA status.

---

### Documentation

- **rbac-setup.md** — clearly written, good migration examples.
- **values.yaml comments** — well-structured, explains `orgAdmin` flag and flow.
- **deploy-rhbk.sh help** — updated with `-f` flag and examples.
- **Missing**: no end-to-end setup sequence. A cluster admin needs to know the exact order: (1) customize `realmUsers` in values, (2) `deploy-rhbk.sh -f values.yaml`, (3) `helm install/upgrade`. The pieces are scattered across files.

---

### Summary

| # | Issue | Severity | Type |
|---|-------|----------|------|
| 1 | CI never provisions viewer user; new tests fail or skip | High | Functional gap |
| 2 | `display_summary()` hardcodes admin info, stale with `-f` | Low | Stale output |
| 3 | `-f` silently ignored if placed after command | Low | UX |
| 4 | `create_realm_users` ignores per-user failure | Medium | Bug |
| 5 | IQE scripts hardcode `username=admin` lookup | Medium | Incomplete refactor |
| 6 | `sync-rbac-admin.sh -f` silent fallback | Medium | Bug |
| 7 | `2>/dev/null` swallows PyYAML error | Low | Bug |
| 8 | Nested bash function definition | Low | Code quality |
| 9 | Default passwords in values.yaml | Low | Security posture |

jordigilh:
Thanks for the thorough review @masayag. Addressed several items in ec31089:

| # | Finding | Status |
|---|---------|--------|
| 1 | CI never provisions viewer user | **Fixed** in a57c205 — `deploy-test-cost-onprem.sh` now passes `-f cost-onprem/values.yaml` to `deploy-rhbk.sh`. CI log confirms 2 users provisioned. |
| 2 | `display_summary()` hardcodes admin info | Ack — low severity, will follow up post-merge. |
| 3 | `-f` silently ignored if placed after command | Ack — documented in help text. Consistent arg parsing across all scripts is a follow-up. |
| 4 | `create_realm_users` ignores per-user failure | **Fixed** in ec31089 — now accumulates a failure counter and returns non-zero with summary. |
| 5 | IQE scripts hardcode `username=admin` | Ack — pre-existing in both `run-iqe-tests.sh` and `run-iqe-tests-local.sh`. Will track separately since it requires IQE fixture changes. |
| 6 | `sync-rbac-admin.sh -f` silent fallback | Ack — valid concern. The script mutating RBAC state should hard-fail when an explicit `-f` parse fails. Will address in follow-up. |
| 7 | `2>/dev/null` swallows PyYAML error | **Fixed** in ec31089 — removed the redirect so stderr propagates. |
| 8 | Nested `_obtain_admin_token` function | Ack — works because bash functions are global, but agree it's fragile. Low severity, follow-up. |
| 9 | Default passwords in values.yaml | Ack — WARNING comment is in place. Pre-GA acceptable. |

**Additional CI fix in ec31089**: The `test_admin_jwt_contains_org_admin_realm_role` was failing because the `_obtain_token()` helper passed `scope: "openid profile email"` which excludes the `roles` client scope. Without `roles` in the scope parameter, Keycloak does not run the `realm_access.roles` mapper. Fixed by adding `roles` to the scope string. The identity propagation tests (`test_admin_gets_full_cost_management_access`) were passing because `conftest.obtain_user_jwt_token_for()` omits the scope parameter entirely, causing all `defaultClientScopes` (including `roles`) to be applied.

jordigilh:
All remaining review items from @masayag are now addressed in the last two commits:

**`451ff994` — `deploy-rhbk.sh` structural improvements:**
- **#2 `display_summary()` hardcodes admin info** — Replaced hardcoded admin/admin block with a dynamic loop that reads the actual provisioned users from `get_realm_users_json()`, showing each user's name, role, and identity attributes.
- **#3 `-f` silently ignored after command** — Merged the two-pass flag+command parser into a single `while` loop so `-f` works regardless of position. Unknown flags now error instead of silently being treated as a command.
- **#8 Nested `_obtain_admin_token`** — Moved to top-level scope with explicit arguments (`keycloak_url`, `username`, `password`), eliminating reliance on the caller's local variables.

**`473870e8` — IQE + sync-rbac-admin hardening:**
- **#5 IQE scripts hardcode `username=admin`** — Both `run-iqe-tests.sh` and `run-iqe-tests-local.sh` now discover the admin username by querying the Keycloak `org-admin` realm-role membership instead of hardcoding `admin`.
- **#6 `sync-rbac-admin.sh -f` silent fallback** — When `-f` is provided, a parse failure or missing `orgAdmin:true` entry is now a hard error (`exit 1`) instead of silently falling back to defaults. Removed `2>/dev/null` that was swallowing stderr from the Python parser.

No remaining technical debt — all items are addressed.

jordigilh:
@masayag — comprehensive status update on all 9 review items:

| # | Finding | Severity | Status | Commit |
|---|---------|----------|--------|--------|
| 1 | CI never provisions viewer user | High | **Fixed** | `a57c205` — `deploy-test-cost-onprem.sh` passes `-f values.yaml`; `fullScopeAllowed: true` on clients |
| 2 | `display_summary()` hardcodes admin info | Low | **Fixed** | `451ff99` — dynamic loop reads actual provisioned users from `get_realm_users_json()` |
| 3 | `-f` silently ignored after command | Low | **Fixed** | `451ff99` — single-pass parser handles flags at any position; unknown flags error |
| 4 | `create_realm_users` ignores per-user failure | Medium | **Fixed** | `ec31089` — tracks `fail_count`, returns non-zero and prints summary on failure |
| 5 | IQE scripts hardcode `username=admin` | Medium | **Fixed** | `473870e` — discovers admin via Keycloak `org-admin` realm-role membership query; `4b77151` — fixed `local` outside function in `run-iqe-tests.sh` |
| 6 | `sync-rbac-admin.sh -f` silent fallback | Medium | **Fixed** | `473870e` — hard-fails (`exit 1`) on parse error or missing `orgAdmin:true` entry |
| 7 | `2>/dev/null` swallows PyYAML error | Low | **Fixed** | `ec31089` — removed `2>/dev/null` from the Python invocation |
| 8 | Nested `_obtain_admin_token` | Low | **Fixed** | `451ff99` — moved to top-level, accepts `keycloak_url`, `username`, `password` as explicit args |
| 9 | Default passwords in values.yaml | Low | **Acknowledged** | WARNING comment already present; acceptable pre-GA. Production Hardening section added to `rbac-setup.md` covering password policy (IA-5) |

All 9 items are resolved. The last CI e2e failure was caused by the `local` keyword used outside a function in `run-iqe-tests.sh` (item #5 fix), which was corrected in `4b77151`.

jordigilh:
### Additional hardening and quality improvements (beyond review feedback)

Beyond the 9 items from your review, we also addressed the following across security, documentation, test quality, Helm templates, and FedRAMP compliance:

**Security**
| ID | Description | Commit |
|---|---|---|
| SEC-01 | Added WARNING comment in `values.yaml` `realmUsers` advising operators to change default passwords before production | [`f2576905`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/f2576905) |
| SEC-02 | `create_realm_users` re-acquires admin token before each user to guard against Keycloak token TTL expiry | [`f2576905`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/f2576905) |
| SEC-03 | Added TLS Note in `deploy-rhbk.sh` header documenting `curl -sk` usage and FedRAMP SC-8 remediation (`--cacert`) | [`f2576905`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/f2576905) |
| SEC-04 | `admin-bootstrap-job.yaml` passed shell variables into a Python heredoc via string interpolation — changed to `os.environ` (same pattern as `sync-rbac-admin.sh`) | [`0175af54`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/0175af54) |
| PRE-01 | Replaced fixed `/tmp/user_response.txt` with `mktemp` to prevent temp file race conditions | [`f2576905`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/f2576905) |
| PRE-09 | Fixed shell injection in `sync-rbac-admin.sh` by passing values as environment variables to `kubectl exec` instead of interpolating into Python code | [`f2576905`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/f2576905) |

**Documentation**
| ID | Description | Commit |
|---|---|---|
| ARCH-03 | Added Breaking Changes section to `rbac-setup.md` — migration instructions for `orgAdminUsernames` removal | [`a45fcc1f`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/a45fcc1f) |
| DOC-01 | Architecture diagram showed hardcoded `is_org_admin=false` — updated to reflect dynamic JWT derivation | [`ecdca6dd`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/ecdca6dd) |
| DOC-02 | Identity Header JSON example showed `is_org_admin: false` — corrected with clarifying note | [`ecdca6dd`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/ecdca6dd) |
| DOC-03 | Troubleshooting curl example hardcoded `is_org_admin:false` — replaced with `IS_ORG_ADMIN` variable | [`ecdca6dd`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/ecdca6dd) |
| SEC-05 | Documented X-Bearer-Token forwarding rationale in Envoy Lua comments | [`ecdca6dd`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/ecdca6dd) |
| PRE-08 | Added inline comment justifying `BYPASS_BOP_VERIFICATION: "True"` in RBAC template | [`a45fcc1f`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/a45fcc1f) |
| FEDRAMP-04 | Added Production Hardening section to `rbac-setup.md` covering AC-7 brute-force protection, IA-5 password policy, SC-8 TLS | [`ecdca6dd`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/ecdca6dd) |

**Test Quality**
| ID | Description | Commit |
|---|---|---|
| QE-02 | Tightened admin access assertion from `any("cost-management" in p)` to `"cost-management:*:*" in permissions` | [`ffce817d`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/ffce817d) |
| PRE-03 | `test_oauth_proxy_no_tls_errors` — switched from `in` to `re.search()` so regex patterns actually work | [`ffce817d`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/ffce817d) |
| TEST-01 | `test_page_loads_without_error` now collects and asserts on error indicators instead of silently printing | [`4c8068e7`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/4c8068e7) |
| TEST-02 | Password grant tests now `pytest.fail` on 4xx/5xx and only `pytest.skip` on `ConnectionError` | [`4c8068e7`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/4c8068e7) |
| TEST-03 | Extracted `decode_jwt_payload` to root `conftest.py` — deduplicated across 4 files | [`4c8068e7`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/4c8068e7) |
| PROD-01 | Added parametrized `test_org_admin_role_determines_access` validating the JWT→RBAC architectural invariant for all user personas | [`4c8068e7`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/4c8068e7) |

**Helm Templates & Architecture**
| ID | Description | Commit |
|---|---|---|
| ARCH-04 | Centralized admin identity — `bootstrapAdmin.username/orgId/accountNumber` now derived from first `orgAdmin:true` in `jwtAuth.realmUsers` via Helm helpers | multiple |
| HELM-01 | Helm bootstrap helpers now use `required` to validate `orgAdmin` entries have `username`, `orgId`, `accountNumber`. `fail` if `bootstrapAdmin.enabled` is true but no `orgAdmin:true` user exists | [`0175af54`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/0175af54) |
| PRE-05 | Password reset in `deploy-rhbk.sh` now checks HTTP status code instead of just shell exit code | [`f2576905`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/f2576905) |
| PRE-07 | Removed duplicate `return 0` from `log_header` function | [`f2576905`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/f2576905) |

**Operational**
| ID | Description | Commit |
|---|---|---|
| SRE-03 | `create_or_update_user` now issues a `DELETE` for the `org-admin` role when `orgAdmin: false`, ensuring role removal on state changes | [`f2576905`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/f2576905) |
| FEDRAMP-03 | Added structured `[AUDIT]` log lines for `assign_role` and `remove_role` actions with timestamp, user, role, realm, and actor | [`f2576905`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/f2576905) |
| `fullScopeAllowed` | Added `fullScopeAllowed: true` to both Keycloak client definitions so realm roles appear in JWT `realm_access.roles` | [`a57c2057`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/a57c2057) |
| `roles` scope | OAuth test token requests now explicitly include `"roles"` in the scope parameter | [`ec31089a`](https://github.com/insights-onprem/cost-onprem-chart/pull/167/commits/ec31089a) |

jordigilh:
@masayag PR to add the provisioning for the admin user is green

masayag (approved):

PR #168, opened 2026-05-11
FLPATH-4184 Add --devel and --chart-version flags for RC/versioned chart testing

testetson22:
- Add `--devel` flag to include pre-release (rc) charts in Helm installation
- Add `--chart-version` flag to pin specific Helm chart versions (e.g., `0.2.19`, `0.3.0-rc1`)
- Enables cleaner RC testing in CI without git checkout workaround

### Test plan

- [x] Verified `--devel --chart-version 0.2.20-rc4` correctly passes `--devel` and `--version` to Helm
- [x] Verified `--chart-version 0.2.19` deploys stable chart successfully
- [x] Verified flag validation rejects `--use-local-chart` combined with `--devel` or `--chart-version`
- [x] Verified CI compatibility - `e2e-commands.sh` already checks for `--devel` support and will use it automatically

### Changes

| File | Changes |
|------|---------|
| `scripts/deploy-test-cost-onprem.sh` | +28 lines - flag parsing, validation, exports, help text |
| `scripts/install-helm-chart.sh` | +7 lines - `USE_HELM_DEVEL` variable and `--devel` flag in helm command |

### Usage

```bash
# Install latest RC chart
./scripts/deploy-test-cost-onprem.sh --devel --namespace cost-onprem

# Install specific version
./scripts/deploy-test-cost-onprem.sh --chart-version 0.2.19 --namespace cost-onprem

# Install specific RC version
./scripts/deploy-test-cost-onprem.sh --devel --chart-version 0.2.20-rc4 --namespace cost-onprem
```

PR #170, opened 2026-05-12
Fix recommend-tests workflow

testetson22:
### Problems

- The GITHUB_TOKEN lacks write permissions on PRs from forks, causing the "Resource not accessible by integration" error when trying to post comments.
- Filter was too strict triggering recommendations of simple changes

##### recommend-tests.yml changes:

- Changed pull_request → pull_request_target (runs with base repo permissions)
- Added step to fetch PR head commit
- Pass PR_HEAD_SHA to the analysis script
##### recommend-tests.sh changes:
- Added PR_HEAD_SHA environment variable support
- Changed git diff commands to use ${PR_HEAD} instead of HEAD

##### test-impact-map.yaml changes:

- deploy-scripts rule: Reduced from extended → smoke profile and medium → low impact
- Pattern broadened to include install, run-pytest, and setup scripts
- Rationale: Script changes (flag additions, help text, etc.) don't typically require deeper IQE testing — smoke tests catch most breakage. This prevents unnecessary test advisor comments on routine script PRs like #168.
- The test advisor will now only recommend deeper testing for high-impact changes: component image updates, Helm template changes, and resource/replica modifications.

**Security note:** This should be safe because the workflow only runs scripts from the base branch (not the PR) and only analyzes the git diff output without executing any code from the PR. If not deemed safe, we'll design another solution.

We can't test this fix without merging to main, but it should be benign, if not a solution for the problem.

### DIFF
--- a/.github/workflows/recommend-tests.yml
+++ b/.github/workflows/recommend-tests.yml
@@ -4,11 +4,15 @@
 # Only comments when changes map to extended/stable/full profiles.
 # Smoke-only changes (UI, init containers, auth proxy) are silent.
 # Uses [TEST-ADVISOR] prefix for idempotent comment updates.
+#
+# NOTE: Uses pull_request_target (not pull_request) to enable comment posting
+# on PRs from forks. This is safe because we only run scripts from the base
+# branch and only analyze the PR diff (no code execution from the PR).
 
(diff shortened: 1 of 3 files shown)


ELK4N4 (approved):
/lgtm

PR #173, opened 2026-05-18
test(rbac): gateway JWT, security boundaries, and lab Keycloak hardening

y-first:
## Summary

Adds chart tests that exercise RBAC through the public API gateway with real Keycloak JWTs (Envoy → Koku/ROS → insights-rbac), complementing existing in-cluster `X-Rh-Identity` coverage in `test_rbac_access.py`.

Adds to existing tests added in [FLPATH-4073](https://redhat.atlassian.net/browse/FLPATH-4073)

- **Gateway RBAC tests** (`test_rbac_gateway.py`): unauthenticated/403 paths, IAM reader vs writer, ROS/Koku denial for unassigned users, migration job presence, and security-boundary cases (tenant isolation, fail-closed, concurrent sessions).
- **E2E password-grant path** (`test_rbac_access.py`): alice/bob/carol gateway JWT tests aligned with RBAC principals.
- **Keycloak helpers** (`rbac_keycloak_users.py`): provision gateway persona users; session sync for `admin`/`viewer` and `org-admin` realm role; attach `roles` client scope to `cost-management-ui` via Admin API (avoids `invalid_scope` when requesting `roles` in the OAuth scope string).
- **Lab resilience**: retry `register_source` on transient HTTP 424 when RBAC is briefly unavailable; `OIDC_PASSWORD_GRANT_SCOPE` for password grant; IAM bootstrap fallback when `rbac:group:read` is not seeded.
- **Docs**: `RBAC_SECURITY_TESTS.md` describes coverage and gaps.

## Commits

1. `test(rbac): add gateway JWT and ops checks for RBAC flows`
2. `test(rbac): cover IAM management API via gateway JWT`
3. `fix(rbac): IAM bootstrap for images without rbac:group:read`
4. `test(rbac): harden gateway JWT suite for lab Keycloak and flaky RBAC`

## Test plan

- [x] Full chart suite on lab cluster (ocp-edge122): **295 passed**, 21 skipped (`deploy-test-cost-onprem.sh --skip-deploy`)
- [ ] CI / recommend-tests workflow on PR
- [ ] Verify Keycloak realm was deployed via `deploy-rhbk.sh` (needs `roles` client scope on `cost-management-ui` for `realm_access.roles` / org-admin assertions)
- [ ] Re-run auth + e2e markers: `pytest -m auth` and `pytest suites/e2e/test_rbac_access.py`

### DIFF
--- a/tests/suites/auth/test_rbac_gateway.py
+++ b/tests/suites/auth/test_rbac_gateway.py
@@ -87,7 +87,7 @@ def _provision_rbac_iam_reader_gateway(
     org_id: str,
     rbac_gateway_test_user_password: str,
 ):
-    """Keycloak user + RBAC group binding with minimal ``rbac:group:read`` (or first rbac perm)."""
+    """Keycloak user + RBAC group binding with read-only rbac IAM (group or principal read)."""
     admin_token = fetch_keycloak_master_admin_token(
         keycloak_config.url,
         cluster_config.keycloak_namespace,
@@ -129,7 +129,7 @@ def _provision_rbac_iam_reader_gateway(
         timeout=120,
     )
     out = (result.stdout or "") + (result.stderr or "")
-    if result.returncode != 0 or "no_rbac_permissions_seeded" in out:
+    if result.returncode != 0 or "no_rbac_iam_read_permission" in out:
         pytest.skip(
             f"RBAC IAM bootstrap skipped (returncode={result.returncode}): {out[:600]}"
         )
@@ -242,17 +242,17 @@ class TestRBACGateway:
             f"got {response.status_code}: {response.text[:300]}"
         )
 
-    def test_gateway_rbac_groups_iam_reader_returns_200(
+    def test_gateway_rbac_principals_iam_reader_returns_200(
         self,
         gateway_url: str,
         http_session: requests.Session,
         gateway_rbac_iam_user_jwt,
     ):
-        """User with ``rbac:group:read`` (or equivalent) can list groups via gateway."""
+        """User with minimal ``rbac`` read permission can list principals via gateway."""
         if not _check_gateway_reachable(gateway_url, http_session):
             pytest.skip("Gateway service not available")
 
-        url = f"{gateway_url.rstrip('/')}/rbac/v1/groups/?limit=5"
+        url = f"{gateway_url.rstrip('/')}/rbac/v1/principals/?limit=5"
         response = http_session.get(
             url,
             headers=gateway_rbac_iam_user_jwt.authorization_header,
@@ -260,12 +260,12 @@ class TestRBACGateway:
             verify=False,
         )
         assert response.status_code == 200, (
-            f"Expected 200 listing groups with IAM read role, "
+            f"Expected 200 listing principals with IAM read role, "
             f"got {response.status_code}: {response.text[:400]}"
         )
         payload = response.json()
         assert "data" in payload or "meta" in payload, (
-            f"Unexpected RBAC groups JSON shape: {list(payload.keys())[:10]}"
+            f"Unexpected RBAC principals JSON shape: {list(payload.keys())[:10]}"
         )
 
     def test_gateway_rbac_groups_post_iam_reader_forbidden(
--- a/tests/suites/auth/test_rbac_gateway.py
+++ b/tests/suites/auth/test_rbac_gateway.py
@@ -355,3 +355,587 @@ def test_rbac_migration_job_completed(cluster_config):
     assert status.get("succeeded") == 1, (
         f"Expected rbac-migration job succeeded=1, got: {json.dumps(status)[:800]}"
     )
+
+
+@pytest.mark.auth
+@pytest.mark.integration
+class TestRBACSecurityBoundaries:
+    """Security boundary tests for RBAC authorization enforcement.
+
+    These tests validate fail-closed behavior, token validation, and tenant
+    isolation to prevent authorization bypass and data leakage.
+    """
+
+    def test_rbac_service_unavailable_denies_access_fail_closed(
+        self,
+        cluster_config,
+        gateway_url: str,
+        http_session: requests.Session,
+        gateway_rbac_iam_user_jwt,
+    ):
+        """CRITICAL: When RBAC API is down, requests MUST fail-closed (deny access).
+
+        This validates that the system fails securely when the authorization
+        backend is unreachable. A 424 (Failed Dependency) or 503 is acceptable
+        only if it denies data access. A 200 with data would be a P0 security bug.
+        """
+        from utils import run_oc_command
+
+        if not _check_gateway_reachable(gateway_url, http_session):
+            pytest.skip("Gateway service not available")
+
+        # Scale RBAC deployment to 0 to simulate service outage
+        scale_result = run_oc_command(
+            [
+                "scale",
+                "deployment",
+                "-n",
+                cluster_config.namespace,
+                "-l",
+                "app.kubernetes.io/component=rbac-api",
+                "--replicas=0",
+            ],
+            check=False,
+        )
+        if scale_result.returncode != 0:
+            pytest.skip(f"Could not scale RBAC deployment: {scale_result.stderr}")
+
+        try:
+            # Wait for pods to terminate
+            import time
+            time.sleep(10)
+
+            # Attempt authenticated request - should fail-closed
+            url = f"{gateway_url.rstrip('/')}/cost-management/v1/reports/openshift/costs/"
+            response = http_session.get(
+                url,
+                headers=gateway_rbac_iam_user_jwt.authorization_header,
+                timeout=60,
+                verify=False,
+            )
+
+            # CRITICAL: Must NOT return 200 with data
+            assert response.status_code != 200, (
+                f"SECURITY VIOLATION: RBAC service down but request succeeded "
+                f"with status {response.status_code}. This is fail-open behavior "
+                f"and exposes data without authorization checks. Response: {response.text[:400]}"
+            )
+
+            # Expected: 424 (Failed Dependency), 503 (Service Unavailable), or 403 (Forbidden)
+            assert response.status_code in (403, 424, 503, 504), (
+                f"Expected fail-closed response (403/424/503/504) when RBAC down, "
+                f"got {response.status_code}: {response.text[:300]}"
+            )
+
+        finally:
+            # Restore RBAC service
+            run_oc_command(
+                [
+                    "scale",
+                    "deployment",
+                    "-n",
+                    cluster_config.namespace,
+                    "-l",
+                    "app.kubernetes.io/component=rbac-api",
+                    "--replicas=1",
+                ],
+                check=False,
+            )
+            # Wait for RBAC to be ready again
+            import time
+            time.sleep(15)
+
+    def test_expired_jwt_rejected(
+        self,
+        gateway_url: str,
+        http_session: requests.Session,
+        keycloak_config,
+        cluster_config,
+        rbac_gateway_test_user_password: str,
+        _provision_rbac_iam_reader_gateway,
+    ):
+        """Expired JWTs (exp claim in the past) MUST be rejected with 401.
+
+        Tests that the gateway enforces JWT expiration to prevent session
+        hijacking with stale credentials.
+        """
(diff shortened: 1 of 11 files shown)


tests/suites/auth/test_rbac_gateway.py
```diff
@@ -0,0 +1,941 @@
...
+from utils import exec_in_pod_raw, get_pod_by_label, get_route_url, run_oc_command
+
+
+RBAC_IAM_GATEWAY_USERNAME = "rbac-iam-admin"
+
+
+@pytest.fixture(scope="session")
+def rbac_gateway_test_user_password() -> str:
+    """Password for synthetic RBAC gateway test users (Keycloak + password grant)."""
+    return os.environ.get("RBAC_GATEWAY_TEST_USER_PASSWORD", "RbacGwTest1!")
```
jordigilh: **[SEC-1 · P0]** Hardcoded password fallback.

`"RbacGwTest1!"` is created on real Keycloak instances when `RBAC_GATEWAY_TEST_USER_PASSWORD` is unset. On non-ephemeral clusters this leaves a known credential.

**Fix**: Remove the default. Either `pytest.skip()` when unset, or generate a random password per-session:
```python
import secrets
return os.environ.get("RBAC_GATEWAY_TEST_USER_PASSWORD") or pytest.skip(
    "RBAC_GATEWAY_TEST_USER_PASSWORD env var required"
)
```

Same issue in `tests/suites/e2e/test_rbac_access.py:556`.

jordigilh (commented):
## GA Readiness Audit — QE Dimensions

Comprehensive audit of RBAC gateway test coverage relative to features deployed in PR #146 (insights-rbac v1) and PR #167 (org-admin from Keycloak realm role). Inline comments below for each finding that needs addressing.

**Summary**: 8 merge-blocker findings (1 P0, 7 P1), plus 1 hardening item (P2) flagged inline.

Full audit canvas available in the workspace at `canvases/pr-173-rbac-ga-audit-v2.canvas.tsx`.

tests/suites/auth/test_rbac_gateway.py
```diff
@@ -0,0 +1,941 @@
...
+        assert response_fresh.status_code == 200, (
+            f"Fresh token should work, got {response_fresh.status_code}"
+        )
+
+        # Document expected behavior for expired tokens
+        # (Full test requires either: time-travel, token forging, or waiting 5+ minutes)
+        # For now, we validate the gateway is configured to check exp claim
+        # by confirming Envoy JWT filter is present in the config
+
+        pytest.skip(
```
jordigilh: **[SKIP-2 · P1]** This test is implementable now with `python-jose` — no need to defer.

Add `python-jose[cryptography]` as a test dependency and create a `forge_jwt()` helper:
```python
from jose import jwt
from cryptography.hazmat.primitives.asymmetric import rsa
from cryptography.hazmat.backends import default_backend
import time

def forge_jwt(claims: dict) -> str:
    key = rsa.generate_private_key(65537, 2048, default_backend())
    return jwt.encode(claims, key, algorithm="RS256")
```

Then replace the skip with:
```python
expired_jwt = forge_jwt({"sub": "test", "exp": int(time.time()) - 3600, "org_id": "test"})
response = http_session.get(url, headers={"Authorization": f"Bearer {expired_jwt}"})
assert response.status_code == 401
```

Envoy rejects because (a) the signature doesn't match the JWKS, and/or (b) the token is expired. Either way, 401 is the correct assertion — the test validates that Envoy's `jwt_authn` filter is configured and active.

tests/suites/auth/test_rbac_gateway.py
```diff
@@ -0,0 +1,941 @@
...
+            pytest.skip("Gateway service not available")
+
+        # This test would require forging JWTs with missing claims
+        # For now, document expected behavior and verify with valid token
+        # Real implementation requires:
+        # - OpenSSL to generate RSA key pair
+        # - Crafting JWT with missing org_id or account_number
+        # - Signing with fake key
+
+        pytest.skip(
```
jordigilh: **[SKIP-3 · P1]** Same as SKIP-2 — implementable now with `forge_jwt()`.

```python
no_sub_jwt = forge_jwt({"org_id": "test", "exp": int(time.time()) + 300})
response = http_session.get(url, headers={"Authorization": f"Bearer {no_sub_jwt}"})
assert response.status_code == 401
```

Envoy rejects the forged signature before the Lua filter even runs, which is the correct security behavior to validate.

tests/suites/auth/test_rbac_gateway.py
```diff
@@ -0,0 +1,941 @@
...
+            verify=False,
+        )
+
+        # We can't inject malicious org_id through the JWT path without forging tokens
+        # This test documents the expected validation behavior
+        # Real implementation would require:
+        # - Direct POST to Koku with crafted X-Rh-Identity header (bypassing gateway)
+        # - OR forging JWTs with malicious org_id claims
+
+        pytest.skip(
```
jordigilh: **[SKIP-4 · P1]** The org_id boundary tests are implementable **with existing PR infrastructure** — no forging needed for most cases.

`ensure_realm_user_with_password()` already accepts an arbitrary `org_id`. Create Keycloak users with malicious org_id values, obtain real (validly-signed) JWTs, and send through the gateway:

```python
ensure_realm_user_with_password(
    org_id='1" OR "1"="1',  # SQL injection
    username="sqli-test", password=pw, ...
)
tok = obtain_password_grant_token("sqli-test", pw, ...)
response = http_session.get(koku_url, headers=tok.authorization_header)
assert response.status_code in (400, 403)  # must reject
```

Envoy will accept these JWTs (valid Keycloak signature), the Lua filter will embed the malicious org_id into `X-Rh-Identity`, and then Koku/RBAC must reject it. This exercises the real defense-in-depth path.

At minimum, unblock the SQL injection, empty, and path traversal cases (P1). Overflow and wrong-tenant can be P2.

tests/suites/auth/test_rbac_gateway.py
```diff
@@ -0,0 +1,941 @@
...
+            f"org_id boundary test for '{malicious_org_id}' requires either: "
+            f"(a) direct in-cluster access to bypass Envoy, or "
+            f"(b) forging JWTs with malicious claims. "
+            f"Expected behavior: system MUST reject with {expected_status}. "
+            f"TODO: Implement via in-cluster pod with crafted X-Rh-Identity header."
+        )
+
+    def test_permission_revocation_honored_after_cache_clear(
```
jordigilh, line 572: **[AQ-2 · P1]** This test removes the user from ALL RBAC groups but never restores membership.

The module-scoped `_provision_rbac_iam_reader_gateway` fixture won't re-run, so any test that runs after this one in the same module and depends on the IAM reader role will fail.

**Fix**: Add a `finally` block that re-runs the bootstrap:
```python
finally:
    script = render_rbac_iam_reader_bootstrap_script(org_id, RBAC_IAM_GATEWAY_USERNAME)
    exec_in_pod_raw(cluster_config.namespace, rbac_pod, [
        "python", "/opt/rbac/rbac/manage.py", "shell", "-c", script
    ], timeout=120)
```

tests/suites/auth/test_rbac_gateway.py
```diff
@@ -0,0 +1,941 @@
...
+            json={"name": "pytest-rbac-iam-write-deny"},
+            timeout=60,
+            verify=False,
+        )
+        assert response.status_code in (403, 400), (
+            f"Expected 403 (forbidden) or 400 (validation before authz) for POST "
+            f"groups without write, got {response.status_code}: {response.text[:400]}"
+        )
+
+    def test_gateway_ros_recommendations_user_without_rbac_returns_403(
```
jordigilh, line 317: **[GAP-1 · P1]** Missing: unauthenticated 401 test for ROS `/recommendations`.

Koku (`/costs/`), RBAC (`/status/`, `/groups/`) all have unauthenticated rejection tests in this class. ROS only has the 403 path (`test_gateway_ros_recommendations_user_without_rbac_returns_403`). If Envoy routing for ROS skipped JWT validation, this gap would not catch it.

**Fix**: Add `test_gateway_ros_recommendations_unauthenticated_returns_401` here, same pattern as the other 401 tests:
```python
def test_gateway_ros_recommendations_unauthenticated_returns_401(
    self, gateway_url, http_session
):
    if not _check_gateway_reachable(gateway_url, http_session):
        pytest.skip("Gateway not available")
    url = f"{gateway_url.rstrip('/')}/cost-management/v1/recommendations/openshift"
    response = http_session.get(url, timeout=20)
    assert response.status_code == 401
```

tests/suites/e2e/test_rbac_access.py
```diff
@@ -760,3 +790,166 @@ def test_alice_payment_spans_two_clusters(
...
+            f"Alice (gateway) got {response.status_code}: {response.text[:400]}"
+        )
+        data = response.json()
+        projects_seen = set()
+        for day in data.get("data", []):
+            for project_group in day.get("projects", []):
+                proj = project_group.get("project")
+                if proj:
+                    projects_seen.add(proj)
+        if projects_seen:
```
jordigilh: **[AQ-1 · P1]** Vacuous assertion — this test silently passes when no data is returned.

If data ingestion failed or the RBAC fixture didn't seed data, `projects_seen` is empty, the `if` guard skips the assertion, and the test passes with 0 assertions fired. Same issue at lines ~878 and ~921 for `clusters_seen` in bob/carol tests.

**Fix**: Remove the `if` guard and assert data exists first:
```python
assert len(projects_seen) > 0, (
    "Alice gateway returned 200 but no project data — "
    "RBAC fixture may not have seeded data"
)
assert projects_seen == {"payment"}, ...
```

The RBAC test fixture guarantees data exists for all three clusters, so empty results indicate a real problem that should fail the test.

tests/suites/auth/test_rbac_gateway.py
```diff
@@ -0,0 +1,941 @@
...
+        Transient 503s under burst are retried; success bar is modest for CI/lab.
+        """
+        if not _check_gateway_reachable(gateway_url, http_session):
+            pytest.skip("Gateway service not available")
+
+        import concurrent.futures
+        import time
+
+        num_concurrent = 6
+        min_success = 2
```
jordigilh: **[AQ-3 · P2]** `min_success = 2` out of 6 concurrent requests (33%) is too lenient — the test passes even if 4 requests fail, proving little about concurrency safety.

**Suggest**: Raise to `min_success = 4` (67%). If lab infra genuinely can't handle 6 concurrent RBAC calls, document the constraint and file a capacity ticket rather than weakening the assertion.

tests/suites/auth/test_rbac_gateway.py
y-first: Fixed

tests/suites/auth/test_rbac_gateway.py
y-first: fixed

tests/suites/auth/test_rbac_gateway.py
y-first: fixed

tests/suites/auth/test_rbac_gateway.py
y-first: fixed

tests/suites/auth/test_rbac_gateway.py
y-first: fixed

tests/suites/auth/test_rbac_gateway.py
y-first: fixed

tests/suites/e2e/test_rbac_access.py
y-first: fixed

tests/suites/auth/test_rbac_gateway.py
y-first: Implemented in 2be8783 as test_org_id_tenant_isolation_boundary_cases: per-case Keycloak user via ensure_realm_user_with_password(org_id=…), password-grant JWT, gateway GET openshift costs. Covers empty, path-traversal, sql-injection (P1) plus overflow and wrong-tenant (P2). Allows 400/401/403/424 (Envoy returns 401 for some malformed org_id values); always asserts != 200. Verified on lab cluster.

jordigilh (approved):
/lgtm

PR #178, opened 2026-05-27
feat: Keycloak-to-RBAC realm-scoped user/group sync CronJob (FLPATH-4294)

jordigilh:
## Summary

Adds a Kubernetes CronJob that periodically synchronizes all users from a Keycloak (RHBK) realm into insights-rbac, replacing the manual `sync-rbac-admin.sh` script and filling the on-prem gap left by the SaaS BOP/mbop + UMB pipeline.

### What's included

- **Python sync script** (`cost-onprem/scripts/sync_keycloak_principals.py`) — Authenticates to Keycloak Admin API via service-account client credentials, paginates through all realm users, upserts `Tenant`/`Principal`/`Group` records in RBAC, maps Keycloak `org-admin` role to the RBAC admin default group, prunes orphaned principals, and logs per-principal audit lines (FedRAMP SI-4 / AC-2(4)).
- **Helm templates** — `CronJob`, `ConfigMap` (script mount), and helper templates gated behind `rbac.keycloakSync.enabled`.
- **Keycloak client provisioning** (`deploy-rhbk.sh`) — Creates the `rbac-keycloak-sync` service-account client in the `KeycloakRealmImport` CR, assigns `realm-management` roles (`view-users`, `query-users`, `view-realm`), extracts and mirrors the client secret to the cost-mgmt namespace.
- **Install flow integration** (`install-helm-chart.sh`) — `trigger_keycloak_sync()` creates a one-off Job from the CronJob after pod readiness and waits for completion, ensuring users are synced before health checks.
- **Removes `scripts/sync-rbac-admin.sh`** — Superseded by the CronJob.
- **Documentation** — Updated `installation.md` and `NOTES.txt` with enable/trigger/troubleshooting instructions.
- **Automated tests** — 14 Helm template tests (CronJob rendering, env vars, guard clauses, feature gating) and 15 Python unit tests (KeycloakClient pagination, token refresh, error handling, input validation).

### FedRAMP & hardening highlights

- Disabled Keycloak users are skipped (AC-2)
- Per-principal structured audit logging with quoted fields (SI-4)
- Pod `seccompProfile: RuntimeDefault`, `runAsNonRoot`, `allowPrivilegeEscalation: false`
- Database operations wrapped in `transaction.atomic()`
- Token expiry tracking with re-authentication at 80% TTL
- Startup readiness checks for RBAC migration dependencies
- `startingDeadlineSeconds`, `concurrencyPolicy: Forbid`
- Helm guard fails render if `keycloakSync.enabled` but no `orgAdmin` user configured

### Known limitations & future improvements

**Observability gap (CronJob architecture):** The CronJob is a batch workload with no persistent metrics endpoint. Prometheus cannot scrape ephemeral job pods, and `automountServiceAccountToken: false` (security hardening) prevents the script from emitting Kubernetes Events. Sync failures are visible only through pod logs and kube-state-metrics' `kube_job_status_failed` metric (if kube-state-metrics is deployed and accessible).

**Future improvement:** Replace the CronJob with a long-running controller/reconciler (analogous to the SaaS BOP service) that exposes `/metrics`, `/healthz`, and emits proper Kubernetes Events — providing first-class observability without depending on external monitoring infrastructure. This would also enable:
- Webhook-driven sync on Keycloak user events (vs polling)
- Kubernetes Event reporting for failure surfacing via `oc get events`
- Health check integration with the install script and readiness probes

**Single-tenant limitation:** The sync targets a single `org_id` (derived from the first `orgAdmin:true` entry in `jwtAuth.realmUsers`). Multi-org deployments would require one CronJob per org (not yet supported). This matches the typical on-prem single-organization model.

## Test plan

- [x] `helm lint cost-onprem/` passes
- [x] `helm template` with `keycloakSync.enabled=true` renders CronJob, ConfigMap correctly
- [x] `helm template` with `keycloakSync.enabled=false` omits sync resources
- [x] Guard clause: render fails when `clientSecretRef.name` is empty
- [x] Guard clause: render fails when no `orgAdmin:true` user exists
- [x] CronJob spec validated (schedule, concurrencyPolicy, backoffLimit, activeDeadlineSeconds)
- [x] Container env vars validated (all 9 required + secretKeyRef)
- [x] Volume mount validated (readOnly ConfigMap with subPath)
- [x] Python KeycloakClient: auth, token freshness, lazy re-auth, pagination, 404 handling
- [x] Python main(): missing env vars exit(1), auth failure exit(1)
- [ ] `deploy-rhbk.sh` provisions the `rbac-keycloak-sync` client and mirrors secret
- [ ] `install-helm-chart.sh` triggers initial sync and waits for completion
- [ ] CronJob syncs users from Keycloak realm into RBAC on schedule
- [ ] Disabled Keycloak users are skipped
- [ ] Orphaned RBAC principals are pruned when `pruneOrphans: true`
- [ ] Audit log lines appear for create, update, skip_disabled, and prune actions


masayag:
@ELK4N4 pls review

jordigilh:
@ELK4N4 PTAL

cost-onprem/templates/NOTES.txt
```diff
@@ -75,6 +75,26 @@ Storage Classes:
...
 
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+{{- if .Values.rbac.keycloakSync.enabled }}
+
+🔄 KEYCLOAK-TO-RBAC USER SYNC
+
+CronJob: {{ include "cost-onprem.rbac.keycloakSync.name" . }}
+Schedule: {{ .Values.rbac.keycloakSync.schedule }}
+Client ID: {{ .Values.rbac.keycloakSync.clientId }}
+Secret: {{ .Values.rbac.keycloakSync.clientSecretRef.name }}
```
ELK4N4: Is it safe to print it?

cost-onprem/scripts/sync_keycloak_principals.py
```diff
@@ -0,0 +1,380 @@
...
+    datefmt="%Y-%m-%dT%H:%M:%S",
+)
+log = logging.getLogger("keycloak-sync")
+
+PAGE_SIZE = 100
+REQUEST_TIMEOUT = 30
+TOKEN_REFRESH_MARGIN = 0.8
+
+
+class KeycloakClient:
```
ELK4N4, line 33: There is no Keycloak SDK for Python?

cost-onprem/scripts/sync_keycloak_principals.py
```diff
@@ -0,0 +1,380 @@
...
+        org_id: Organization identifier for the RBAC tenant.
+        account_number: Account number for tenant naming.
+        kc_users: Pre-fetched list of Keycloak user dicts for this org.
+        admin_usernames: Set of usernames that should have admin access.
+        prune_orphans: Whether to delete RBAC principals absent from kc_users.
+    """
+    from api.models import Tenant
+    from django.core.management import call_command
+    from django.db import transaction
+    from management.models import Group, Policy, Principal, Role
```
ELK4N4, line 151: Imports should be at the top of the file

cost-onprem/scripts/sync_keycloak_principals.py
```diff
@@ -0,0 +1,380 @@
...
+        group_id = group["id"]
+        group_name = group.get("name", "")
+        attrs = group.get("attributes", {})
+
+        org_id_list = attrs.get("org_id", [])
+        acct_list = attrs.get("account_number", [])
+        org_id = org_id_list[0] if org_id_list else ""
+        account_number = acct_list[0] if acct_list else ""
+
+        if not org_id or not account_number:
```
ELK4N4, line 302: I would use `None` for no results

cost-onprem/scripts/sync_keycloak_principals.py
```diff
@@ -0,0 +1,380 @@
...
+            else:
+                log.warning("[%s] No '%s' sub-group found; no users will be org-admin", org_id, org_admin_subgroup)
+        except Exception:
+            log.exception("[%s] Failed to fetch admin sub-group members", org_id)
+
+        ok = sync(org_id, account_number, members, admin_usernames, prune_orphans)
+        if not ok:
+            all_ok = False
+
+    from django.core.cache import cache
```
ELK4N4, line 337: Imports should be at the top

ELK4N4 (changes requested):

cost-onprem/templates/NOTES.txt
jordigilh: Good catch on the ambiguous label. This prints the K8s Secret **resource name** (e.g. `keycloak-client-secret-rbac-sync`), not the credential value itself. Renamed the label to "Client Secret K8s Secret" to make this clear. Fixed in 6d709fd.

cost-onprem/scripts/sync_keycloak_principals.py
jordigilh: Intentionally using stdlib `urllib` to keep the script zero-dependency. The script runs inside the RBAC container image via `manage.py shell`, so we can only rely on packages already installed in that image. Adding `python-keycloak` would mean either baking it into the RBAC image (upstream change) or adding a `pip install` step to the CronJob init (fragile, slow, requires network). The Keycloak Admin REST API surface we need is small (token + paginated GET), so raw urllib is sufficient.

cost-onprem/scripts/sync_keycloak_principals.py
jordigilh: These imports are intentionally deferred. The script is piped into `manage.py shell` (`python manage.py shell < sync_keycloak_principals.py`), and Django models (`api.models.Tenant`, `management.models.*`) aren't available at module parse time — only after Django's app registry is initialized by the shell command. Moving them to the top would cause `ImportError: Apps aren't loaded yet` or `django.core.exceptions.AppRegistryNotReady`.

cost-onprem/scripts/sync_keycloak_principals.py
jordigilh: Good point. Changed to `None` in 6d709fd. The downstream `if not org_id or not account_number` guard handles both `None` and empty string, so behavior is unchanged.

cost-onprem/scripts/sync_keycloak_principals.py
jordigilh: Same reason as the other deferred imports above — `django.core.cache` requires Django's app registry to be fully initialized, which only happens inside `manage.py shell`. Moving it to module level would raise `AppRegistryNotReady`.

ELK4N4 (approved):
/lgtm

PR #182, opened 2026-06-16
FLPATH-4091 Template or split out various chart config 

testetson22:
This is the first of a few PRs that enables and/or enhances performance testing with larger scale profiles. If the chart persists after the operator is done, this will be generally useful for tuning a deployment, I'd expect. 

Verbose findings, rightsizing, and issues will follow in subsequent PRs.

- **Template Envoy gateway configuration**: Remove the redundant global `request_timeout` (which silently capped all requests at 60s, conflicting with `stream_idle_timeout: 300s`), and make the ingress route `timeout` and `per_try_timeout` configurable via `jwtAuth.envoy.ingressTimeout` / `ingressPerTryTimeout` in values.yaml. Also template the gateway replica count (was hardcoded to 2).
- **Separate ROS processor resources**: Give the ROS processor its own `resources.rosProcessor` block instead of sharing `resources.application` with lighter services. The processor's Go-based workload has different memory characteristics than the Python API services.
- **Template replica counts**: Make Kruize (`kruize.replicas`) and ROS processor (`ros.processor.replicas`) replica counts overridable at deploy time for performance tuning.

## Test plan

- [x] `helm template` renders correctly with default values (all defaults preserved)
- [x] `helm template` renders correctly with overridden values (e.g., `--set jwtAuth.envoy.ingressTimeout=120s`)
- [x] Deploy to test cluster and verify gateway, Kruize, and ROS processor pods start normally
- [x] Run chart test suite: `./scripts/run-pytest.sh --helm`
- [x] Run E2E suite: `./scripts/run-pytest.sh --e2e`

Ref: FLPATH-4091

### DIFF
--- a/cost-onprem/templates/gateway/deployment.yaml
+++ b/cost-onprem/templates/gateway/deployment.yaml
@@ -7,7 +7,7 @@ metadata:
     {{- include "cost-onprem.labels" . | nindent 4 }}
     app.kubernetes.io/component: gateway
 spec:
-  replicas: {{ .Values.jwtAuth.envoy.replicas | default 1 }}
+  replicas: {{ .Values.jwtAuth.envoy.replicas | default 2 }}
   selector:
     matchLabels:
       {{- include "cost-onprem.selectorLabels" . | nindent 6 }}
--- a/cost-onprem/templates/gateway/deployment.yaml
+++ b/cost-onprem/templates/gateway/deployment.yaml
@@ -7,7 +7,7 @@ metadata:
     {{- include "cost-onprem.labels" . | nindent 4 }}
     app.kubernetes.io/component: gateway
 spec:
-  replicas: {{ .Values.jwtAuth.envoy.replicas | default 2 }}
+  replicas: {{ .Values.jwtAuth.envoy.replicas }}
   selector:
     matchLabels:
       {{- include "cost-onprem.selectorLabels" . | nindent 6 }}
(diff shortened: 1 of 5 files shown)


cost-onprem/templates/gateway/deployment.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     {{- include "cost-onprem.labels" . | nindent 4 }}
     app.kubernetes.io/component: gateway
 spec:
-  replicas: 2
+  replicas: {{ .Values.jwtAuth.envoy.replicas | default 2 }}
```
chadcrum: Should-fix: `default` treats `0` as falsy in Go templates, so setting `replicas: 0` in values.yaml silently becomes `replicas: 2`. That breaks intentional scale-to-zero (useful for perf testing isolation).

Since these defaults are already declared in values.yaml, the `| default N` is redundant — the costManagement deployment templates already reference `.Values.costManagement.api.replicas` without it. I'd drop the `| default` here and on the kruize and ros-processor replica lines to match that pattern.

chadcrum (dismissed):
Clean separation of the chart config — templating the timeouts and replica counts makes per-environment tuning much easier without touching the templates. Left one note on the `| default` for replicas potentially blocking scale-to-zero if that's ever a use case.

chadcrum (approved):

PR #183, opened 2026-06-16
COST-7732 Deploy script refactor

testetson22:
This is the second of a few PRs that enables and/or enhances performance testing with larger scale profiles. This is focused on refactoring the relevant scripts for better composability and to partially deconstruct the monolithic `deploy-test-cost-onprem.sh`. 

**This was branched from a longer running branch that expanded in scope too far, thus it will reference some scripts conditionally that may not exist yet.**

- **Refactor `deploy-test-cost-onprem.sh`**: Overhaul flag parsing to support `--perf-only` (skip deploy, run only perf tests), `--perf-suite` (target specific suites like `ros,ingestion`), and observability flags (`--deploy-observability`, `--collect-metrics`). Default `--listener-cpu` to `max` for performance runs since the 300m default throttles all ingestion tests. Simplify argument parsing and remove duplicated logic.
- **Extract listener CPU management**: New `scripts/lib/listener-cpu.sh` library with replica-aware CPU calculation that prevents cluster overcommit when multiple listener pods are running (previously caused full stack crashes).
- **Improve supporting scripts**: Add Python/PyYAML auto-detection to `deploy-rhbk.sh`, Kafka readiness helpers to `deploy-kafka.sh`, and `--performance` mode with unified output directories to `run-pytest.sh`.

## Test plan

- [x] `./scripts/deploy-test-cost-onprem.sh --dry-run --verbose` shows correct flag parsing
- [x] `./scripts/deploy-test-cost-onprem.sh --perf-only --perf-profile baseline --dry-run` skips deploy steps
- [x] CI workflow validates new flags: `./scripts/qe/test-gh-workflow-locally.sh .github/workflows/validate-deploy-test-script.yml`
- [x] Full deploy + chart tests pass: `./scripts/deploy-test-cost-onprem.sh --namespace cost-onprem --verbose`

Ref: COST-7732


scripts/deploy-test-cost-onprem.sh
```diff
@@ -1338,6 +1177,11 @@ main() {
     deploy_helm_chart
     setup_tls
```
chadcrum, line 1193: The `deploy_observability` call here fires whenever the function exists — it doesn't check `DEPLOY_OBSERVABILITY`. Once `perf-observability.sh` lands, this will deploy exporters on every run even without `--deploy-observability`.

Something like:
```bash
if [[ "${DEPLOY_OBSERVABILITY}" == "true" ]]; then
    if type deploy_observability &>/dev/null; then
        deploy_observability
    else
        log_warning "--deploy-observability requested but perf-observability.sh not found; skipping"
    fi
fi
```
The flag is already parsed and shown in the summary — just needs to gate the actual call too.

scripts/lib/listener-cpu.sh
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/env bash
+# scripts/lib/listener-cpu.sh — Listener CPU management for test runs
+#
+# Requires: log_info, log_success, log_warning, log_error, log_step,
+#           log_verbose, parse_cpu_to_millicores (all from parent script)
```
chadcrum: nit: `parse_cpu_to_millicores` is listed under "Requires: ... (all from parent script)" but it's defined in this file at line 19. Minor — just remove it from the Requires list.

chadcrum (dismissed):

chadcrum:
`run_performance_tests` is called directly but the function definition was removed in this PR (moving to `perf-testing.sh`). Unlike `deploy_observability` which has a `type` guard, these call sites will crash with "command not found" under `set -e` if someone uses `--run-perf` or `--perf-only` before the companion PR lands.

The CI tests pass because `--dry-run` returns before hitting `run_tests()`, so this won't get caught there.

Could add the same `type` guard pattern used for `deploy_observability`:
```bash
if type run_performance_tests &>/dev/null; then
    run_performance_tests
else
    log_error "Performance testing library not found (scripts/lib/perf-testing.sh)"
    return 1
fi
```

chadcrum (approved):

PR #184, opened 2026-06-16
FLPATH-4061 Test Observability Stack

testetson22:
This is the third of a few PRs that enables and/or enhances performance testing with larger scale profiles. This brings in an observability stack that is key to measuring performance tests.

- **Observability deployment**: New `deploy-observability.sh` that installs celery-exporter, postgres_exporter, valkey-exporter, and Prometheus ServiceMonitors for all cost-onprem components. Integrated via `--deploy-observability` flag in the deploy script.
- **Metrics collection pipeline**: Continuous Prometheus metric snapshots during performance runs, capturing CPU, memory, request rates, and queue depths at configurable intervals.
- **Report generation**: Three report generators — per-run HTML report with charts and KPI analysis, cross-run comparison matrix for trend analysis, and JSON summary for programmatic consumption.
- **S3 integration**: Upload collected metrics and reports to S3-compatible storage for long-term archival and cross-run comparison.

## Test plan

- [x] `./scripts/deploy-observability.sh --dry-run` completes without errors
- [x] `./scripts/observability/collect-metrics.sh --help` shows usage
- [x] Python scripts import successfully: `python scripts/observability/generate-perf-summary.py --help`
- [x] Full integration: `./scripts/deploy-test-cost-onprem.sh --deploy-observability --run-perf --collect-metrics --dry-run`

Ref: FLPATH-4061

depends-on: https://github.com/insights-onprem/cost-onprem-chart/pull/183


scripts/observability/collect-metrics.sh
```diff
@@ -0,0 +1,628 @@
...
+log_success() { echo -e "${GREEN}[SUCCESS]${NC} $1"; }
+log_warning() { echo -e "${YELLOW}[WARNING]${NC} $1"; }
+log_error() { echo -e "${RED}[ERROR]${NC} $1" >&2; }
+
+# Generate TEST_RUN_ID in format: {chart_version}-{perf_profile}-{epoch_time}
+generate_test_run_id() {
+    if [[ -n "$TEST_RUN_ID" ]]; then
+        log_info "Using provided TEST_RUN_ID: $TEST_RUN_ID"
+        return 0
+    fi
```
chadcrum, line 90: nit: `auth_header` is computed here but never used — the curl call below passes `-H "Authorization: Bearer $token"` directly. When `$token` is empty (e.g., cert-based auth), this sends an empty `Bearer` header instead of omitting it. Probably fine in practice but the dead variable is confusing.

Something like `${token:+-H "Authorization: Bearer $token"}` in the curl call would clean it up.

scripts/observability/generate-perf-summary.py
```diff
@@ -0,0 +1,568 @@
...
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, Optional
+import xml.etree.ElementTree as ET
+
+
+# ---------------------------------------------------------------------------
+# Data loading
+# ---------------------------------------------------------------------------
+
```
chadcrum, line 39: Not blocking, but `load_session()`, `load_metadata()`, and `parse_junit()` are duplicated across all four Python scripts with slight variations. Might be worth extracting into a shared module at some point — would make it easier to fix bugs in one place instead of 3-5.

scripts/observability/generate-perf-run-report.py
```diff
@@ -0,0 +1,1755 @@
...
+#   green  : value OP green_threshold
+#   yellow : value OP yellow_threshold  (but not green)
+#   red    : otherwise
+#
+# ``op`` is one of "<" (lower is better, e.g. latency) or ">" (higher is
+# better, e.g. throughput).  For boolean metrics use "==" with 1/0.
+# ---------------------------------------------------------------------------
+
+KPI_THRESHOLDS: dict[str, list[dict]] = {
+    # --- API Latency (P95 targets in seconds) ---
```
chadcrum, line 46: Heads up — the `dict | None` and `list[dict]` type hint syntax requires Python 3.10+. If this ever needs to run on 3.9 (some OCP build images still ship it), adding `from __future__ import annotations` at the top would fix it. Not a problem if 3.10+ is guaranteed.

chadcrum (approved):

y-first (approved):

scripts/observability/generate-perf-summary.py
testetson22: I may have split this out on the larger branch with a bunch of other test fixes, but I'll take care of it in any case in the next PR. Thanks, good catch.

scripts/observability/generate-perf-run-report.py
testetson22: I'll keep an eye out for issues, but this should only even be ran on Jenkins hosts, which I guess could have the same problem.

scripts/observability/collect-metrics.sh
testetson22: I'll clean it up in the next one; might have already, I just gave up back-porting a few changes ago since this is mostly inert until the final PR with the test fixes.

chadcrum (approved):
/lgtm

PR #169, opened 2026-05-12
[FLPATH-3267] Add TLS support for external Valkey/Redis

ELK4N4:
## Summary

- Add TLS configuration for external Valkey/Redis connections (`valkey.tls.enabled`, `valkey.tls.caCertSecretName`)
- Inject `REDIS_SSL` and `REDIS_SSL_CA_CERTS` env vars into all Koku workloads
- Mount CA certificate Secret as a volume when provided
- Add Helm template tests and expand documentation

## Changes

**`cost-onprem/values.yaml`**
- Add `valkey.tls.enabled` (default `false`) and `valkey.tls.caCertSecretName` (default `""`)

**`cost-onprem/templates/_helpers-koku.tpl`**
- Inject `REDIS_SSL=True` env var when `valkey.tls.enabled`
- Inject `REDIS_SSL_CA_CERTS=/etc/redis-tls/ca.crt` when CA cert secret is provided
- Add `redis-tls-ca` volumeMount and volume (Secret) to shared Koku helpers

**`tests/suites/helm/test_chart_lint.py`**
- Add `TestExternalValkey` class with 7 tests: auth, TLS, combined, TLS without CA cert, default no-TLS, auth validation failure

**`docs/operations/configuration.md`**
- Expand External Valkey/Redis section with TLS examples, Secret creation commands, and required Redis configuration guidance

## Test plan

- [ ] `helm template` renders correctly with default values (no TLS env vars)
- [ ] `helm template --set valkey.tls.enabled=true` renders `REDIS_SSL` without volume
- [ ] `helm template --set valkey.tls.enabled=true --set valkey.tls.caCertSecretName=my-ca` renders volume + env
- [ ] `helm template --set valkey.auth.enabled=true --set valkey.auth.secretName=foo --set valkey.tls.enabled=true --set valkey.tls.caCertSecretName=bar` renders both auth and TLS
- [ ] E2E: deploy with external TLS Redis on OpenShift

## Related

- Koku backend counterpart: project-koku/koku#6054
- Jira: [FLPATH-3267](https://redhat.atlassian.net/browse/FLPATH-3267)

PR #186, opened 2026-06-18
[COST-7740] Fix koku migration job missing TLS volume mounts

ELK4N4:
## Summary

- Replace hardcoded `volumeMounts`/`volumes` in the koku migration job with the shared helpers (`cost-onprem.koku.volumeMounts` / `cost-onprem.koku.volumes`)
- Add `cost-onprem.koku.initContainer.combineCA` init container to populate the combined CA bundle
- Ensures `redis-tls-ca` volume is mounted when `valkey.tls.caCertSecretName` is set, matching all other 12 koku workloads

## Root Cause

PR #169 (FLPATH-3267) added TLS volume support to the RBAC migration job but missed the koku migration job. The koku job had hardcoded volumes (`tmp` only) and was never updated to use the shared helpers that were later extended for TLS.

## Impact

When `valkey.tls.caCertSecretName` is set, the `commonEnv` helper injects `REDIS_SSL_CA_CERTS="/etc/redis-tls/ca.crt"` into the job's environment, but the Secret volume containing `ca.crt` was never mounted — the file didn't exist in the container. Currently latent (Django migrations don't touch Redis), but would cause `FileNotFoundError` if any future code adds a cache operation during `django.setup()` or migration execution.

## Test plan

- [x] `helm lint` passes with TLS values enabled
- [x] `helm template` with `--set valkey.tls.enabled=true --set valkey.tls.caCertSecretName=redis-tls-ca` renders `redis-tls-ca` volume + mount in the migration job
- [x] `helm template` without TLS values omits `redis-tls-ca` (conditional logic preserved)
- [x] Deploy to cluster and verify migration job pod has correct volumes

### DIFF
--- a/cost-onprem/templates/cost-management/jobs/migration.yaml
+++ b/cost-onprem/templates/cost-management/jobs/migration.yaml
@@ -26,9 +26,6 @@ spec:
       securityContext:
         {{- include "cost-onprem.securityContext.pod" . | nindent 8 }}
 
-      initContainers:
-      {{- include "cost-onprem.koku.initContainer.combineCA" . | nindent 6 }}
-
       containers:
       - name: migrate
         image: "{{ include "cost-onprem.koku.image" . }}"
@@ -115,7 +112,13 @@ spec:
         {{- include "cost-onprem.koku.commonEnv" . | nindent 8 }}
 
         volumeMounts:
-        {{- include "cost-onprem.koku.volumeMounts" . | nindent 8 }}
+        - name: tmp
+          mountPath: /tmp
+        {{- if and .Values.valkey.tls.enabled .Values.valkey.tls.caCertSecretName }}
+        - name: redis-tls-ca
+          mountPath: /etc/redis-tls
+          readOnly: true
+        {{- end }}
 
         securityContext:
           {{- include "cost-onprem.securityContext.container" . | nindent 10 }}
@@ -129,4 +132,13 @@ spec:
             memory: 1Gi
 
       volumes:
-      {{- include "cost-onprem.koku.volumes" . | nindent 6 }}
+      - name: tmp
+        emptyDir: {}
+      {{- if and .Values.valkey.tls.enabled .Values.valkey.tls.caCertSecretName }}
+      - name: redis-tls-ca
+        secret:
+          secretName: {{ .Values.valkey.tls.caCertSecretName }}
+          items:
+            - key: ca.crt
+              path: ca.crt
+      {{- end }}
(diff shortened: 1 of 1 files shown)


testetson22 (dismissed):
You're likely fine with just the E2E run that's been automatically triggered. The test recommendation flow is meant to be conservative around chart/template changes; you should exercise this path with currently running test.

testetson22:
@ELK4N4 The E2E tests are failing during deployment. The ConfigMaps may need to also be pre-install hooks with hook-weight: "-1" so they exist by the time the Job (weight 0) tries to mount them. 

CMs:
- {{ fullname }}-aws-config — from cost-onprem/templates/cost-management/configmap-aws-config.yaml
- {{ fullname }}-ca-combine — from cost-onprem/templates/cost-management/configmap-ca-combine.yaml
- {{ fullname }}-service-ca — from cost-onprem/templates/shared/configmap-service-ca.yaml

chadcrum (commented):
Nice catch on the missing TLS volume mounts — the migration job definitely needs `redis-tls-ca` when Valkey TLS is enabled.

The issue is that the shared `cost-onprem.koku.volumeMounts` / `cost-onprem.koku.volumes` helpers pull in three ConfigMap volumes (`cost-onprem-aws-config`, `cost-onprem-ca-combine`, `cost-onprem-service-ca`) that are regular chart resources. The migration job is a pre-install hook (weight 0), so on fresh installs it runs *before* those ConfigMaps exist. The pod can't start and Helm times out waiting:

```
Release "cost-onprem" does not exist. Installing it now.
Error: failed pre-install: resource not ready, name: cost-onprem-koku-migrate, kind: Job, status: InProgress
context deadline exceeded
```

CI e2e hit this on a clean install.

The migration job only runs `manage.py migrate` — it talks to PostgreSQL and nothing else. It doesn't need the AWS config, combined CA bundle, or the `combineCA` init container. The only extra volume it needs is `redis-tls-ca` (a user-provided Secret that exists before install) so Django's cache backend can find the CA cert path at setup time.

The RBAC migration job (`templates/rbac/migration-job.yaml`) and admin-bootstrap job already handle this correctly — inline conditional mounts instead of the shared helpers. Something like:

```yaml
        volumeMounts:
        - name: tmp
          mountPath: /tmp
        {{- if and .Values.valkey.tls.enabled .Values.valkey.tls.caCertSecretName }}
        - name: redis-tls-ca
          mountPath: /etc/redis-tls
          readOnly: true
        {{- end }}

      volumes:
      - name: tmp
        emptyDir: {}
      {{- if and .Values.valkey.tls.enabled .Values.valkey.tls.caCertSecretName }}
      - name: redis-tls-ca
        secret:
          secretName: {{ .Values.valkey.tls.caCertSecretName }}
          items:
            - key: ca.crt
              path: ca.crt
      {{- end }}
```

No `initContainers` block needed. This keeps the hook self-contained and avoids the dependency on chart resources that don't exist yet during pre-install.

chadcrum (approved):

PR #191, opened 2026-07-10
test: Add AWS S3 test coverage and fix addressing_style in Helm ConfigMap

chadcrum:
## Summary

- Add 92 pytest tests (87 offline, 5 cluster-required) covering AWS S3 support in `install-helm-chart.sh` and Helm chart templates
- Fix `aws-config` ConfigMap that hardcoded `addressing_style=path`, which breaks AWS S3 (needs `auto` for virtual-hosted style addressing)
- Fix autouse session fixture that skipped all tests (including offline helm/lint) when no cluster was connected
- Fix double-port bug in `create_s3_buckets` URL construction when `S3_ENDPOINT` includes a port

## Bug fixes

**addressing_style hardcoded to path** - The `aws-config` ConfigMap hardcoded `addressing_style=path`, which is correct for on-prem S3 (NooBaa, Ceph RGW) but incorrect for AWS S3. AWS requires `auto` (virtual-hosted style). The install script now detects AWS endpoints and injects `--set objectStorage.s3.addressingStyle=auto` during helm install.

**Double-port in bucket job URL** - When `S3_ENDPOINT` included a port (e.g., `s4.s4.svc.cluster.local:7480`), the URL construction appended the default port again, producing malformed URLs like `https://host:7480:443`. Now uses `endpoint_host` (already stripped by `parse_s3_host`) for URL construction and extracts an embedded port as the fallback before defaulting to 443.

## Test coverage

| Suite | Tests | Requires cluster |
|-------|-------|-----------------|
| `test_install_script.py` (new) | 58 | No |
| `test_chart_lint.py` (extended) | 21 new | No |
| `test_storage.py` (extended) | 5 new | Yes |

**Install script tests** - bash functions tested via `subprocess.run` with `BASH_SOURCE` guard: S3 URL parsing, AWS endpoint detection, bucket name validation, region resolution, SSL auto-enable, `create_s3_buckets` decision logic, `deploy_helm_chart` `--set` injection.

**Helm template tests** - offline `helm template` rendering: ConfigMap values, endpoint URL port handling, credential security (secretKeyRef, no inline creds), bucket name overrides.

**Deployed config tests** - live cluster validation: ConfigMap exists, SigV4 signing, addressing style matches backend type (path for on-prem, auto for AWS).

## Other changes

- `conftest.py`: Fix `_ensure_keycloak_password_grant_lab_users` autouse fixture - detects Keycloak inline instead of depending on fixtures that call `pytest.skip()`
- `conftest.py`: `s3_config` fixture reads from deployed state (works for S4/ODF/AWS)
- `cleanup.py`: Accept `addressing_style` parameter
- `install-helm-chart.sh`: `BASH_SOURCE` guard for sourcing in tests

## Validation

- All 87 offline tests pass without cluster access
- Full pytest suite (427 passed, 0 failed) on AWS S3 backend (us-east-1) on lab cluster
- S4 backward compatibility verified (47 infrastructure tests pass, addressing_style=path, default buckets)
- Manually verified addressing_style=auto on AWS, path on S4/NooBaa
- 8/8 manual test cases pass (MT-10 through MT-17)

### DIFF
--- a/scripts/install-helm-chart.sh
+++ b/scripts/install-helm-chart.sh
@@ -829,13 +829,13 @@ create_s3_buckets() {
             verify_ssl="true"
             echo_info "  Auto-enabled SSL verification for AWS S3 endpoint (override with S3_VERIFY_SSL=false)"
         fi
-        local s3_port="${S3_PORT:-443}"
+        local s3_port="${S3_PORT:-$(echo "$S3_ENDPOINT" | grep -oP ':\K[0-9]+$' || echo 443)}"
         local s3_ssl="${S3_USE_SSL:-true}"
         if [ "$s3_ssl" = "true" ]; then
-            s3_url="https://${S3_ENDPOINT}:${s3_port}"
+            s3_url="https://${endpoint_host}:${s3_port}"
             [ "$verify_ssl" = "true" ] && no_verify_ssl="" || no_verify_ssl="--no-verify-ssl"
         else
-            s3_url="http://${S3_ENDPOINT}:${s3_port}"
+            s3_url="http://${endpoint_host}:${s3_port}"
             no_verify_ssl=""
         fi
         echo_info "  ✓ Using S3_ENDPOINT: $s3_url"
--- a/scripts/install-helm-chart.sh
+++ b/scripts/install-helm-chart.sh
@@ -2315,6 +2315,9 @@ main() {
     echo_info "Next: Run NAMESPACE=$NAMESPACE ./run-pytest.sh to test the deployment"
 }
 
+# Allow sourcing for unit tests: only run main when executed directly.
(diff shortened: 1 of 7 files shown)


y-first (approved):

testetson22 (approved):
Good stuff. I like the source guard, though it does make me consider that `parse_s3_host`, `is_aws_s3_endpoint_host`, `validate_s3_bucket_name`, etc. could be in a shared `libs/s3-utils.sh` script. Not blocking, just a thought/nitpick.

chadcrum:
Fair point on the utility functions. For now though, install-helm-chart.sh is a standalone script users download and run directly - splitting into libs adds a distribution dependency we don't need yet. Worth revisiting if it grows more, but not the tradeoff right now.

PR #192, opened 2026-07-10
fix(helm): decouple Kafka TLS env vars from SASL gate (COST-7893)

chadcrum:
## Summary

Kafka TLS environment variables (KAFKA_SECURITY_PROTOCOL, KAFKA_SSL_CA_LOCATION) were nested inside the SASL gate, so TLS-only deployments (SSL without SASL) never rendered them. Extract into independent tlsEnv/ingressTlsEnv helpers, align the CA location gate with volume/mount requirements (both tls.enabled AND tls.caCertSecret), remove duplicate KAFKA_SECURITY_PROTOCOL from ROS templates, and add Kafka TLS to the migration job.

For rendered before/after YAML showing exactly what was missing and what the fix produces for each affected manifest, see the [detailed Jira comment on COST-7893](https://redhat.atlassian.net/browse/COST-7893?focusedId=17561306&focusedCommentId=17561306#comment-17561306).

## Changes

- Refactored `_helpers.tpl` to separate TLS env vars from SASL logic
- Added `tlsEnv` and `ingressTlsEnv` helpers for consistent CA/security protocol rendering
- Updated all Koku, ROS, and ingress deployments to use the new helpers
- Fixed CA location gating to match volume/mount logic (prevents phantom env vars)
- Removed duplicate KAFKA_SECURITY_PROTOCOL declarations in ROS templates
- Added TLS volume/mount to migration job so it can reach Kafka with certs

## Verification

All four bugs are fixed by the refactored helpers:
1. TLS env vars now render regardless of SASL configuration
2. CA location gate aligned with volume availability
3. No more duplicate env vars in ROS deployments
4. Migration job has TLS support matching other components

### DIFF
--- a/cost-onprem/templates/_helpers-koku.tpl
+++ b/cost-onprem/templates/_helpers-koku.tpl
@@ -339,6 +339,7 @@ Common environment variables for Koku API and Celery
 - name: INSIGHTS_KAFKA_PORT
   value: {{ include "cost-onprem.koku.kafka.port" . | quote }}
 {{- include "cost-onprem.kafka.saslEnv" . }}
+{{- include "cost-onprem.kafka.tlsEnv" . }}
 - name: S3_ENDPOINT
   value: {{ include "cost-onprem.storage.endpointWithProtocol" . | quote }}
 - name: REQUESTED_BUCKET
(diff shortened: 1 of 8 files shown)


y-first (approved):

testetson22 (approved):
/lgtm

ELK4N4 (approved):

PR #232, opened 2026-08-07
docs: operator test compat notes and harness hooks

martinpovolny:
## Summary
- Document how to run chart pytest against `koku-server-operator` deployments (`docs/development/operator-test-compat.md`).
- Add `DEPLOYMENT_MODE=operator` (auto-excludes the helm suite) and `STORAGE_SECRET_NAME` for BYOI/operator Secret discovery.
- Capture gaps found on clusterbot: Ingress/RBAC/UI not deployed by the operator yet; Celery worker label differences; naming that already aligns (`{release}-koku-api`, `{release}-api` route, gateway).

## Test plan
- [ ] With a Ready operator CR named `cost-onprem` in `cost-tests`:
  `NAMESPACE=cost-tests HELM_RELEASE_NAME=cost-onprem DEPLOYMENT_MODE=operator ./scripts/run-pytest.sh --auth --infrastructure --smoke --no-ui`
- [ ] Confirm helm suite is skipped under `DEPLOYMENT_MODE=operator`
- [ ] Confirm default helm mode is unchanged (`DEPLOYMENT_MODE` unset)

### DIFF
--- a/scripts/run-pytest.sh
+++ b/scripts/run-pytest.sh
@@ -47,6 +47,8 @@
 #   NAMESPACE              Target namespace (default: cost-onprem)
 #   HELM_RELEASE_NAME      Helm release name (default: cost-onprem)
 #   KEYCLOAK_NAMESPACE     Keycloak namespace (default: keycloak)
+#   DEPLOYMENT_MODE        helm (default) or operator — excludes helm suite when operator
+#   STORAGE_SECRET_NAME    Optional override for S3 credentials Secret discovery
 #   PYTHON                 Python interpreter (default: python3)
 #
 # Examples:
@@ -535,6 +537,38 @@ main() {
         # Export for run_pytest to use for HTML report
         export PERF_REPORTS_DIR="$perf_reports_dir"
         log_info "Performance test reports will be saved to: ${perf_reports_dir}/"
+    fi
+
+    # Operator-managed deploys: chart lint/template suites do not apply.
+    if [[ "${DEPLOYMENT_MODE:-helm}" == "operator" ]]; then
+        local requested_helm=false
+        for _m in "${pytest_markers[@]+"${pytest_markers[@]}"}"; do
+            if [[ "$_m" == "helm" ]]; then
+                requested_helm=true
+                break
+            fi
+        done
+        if [[ "$requested_helm" != "true" ]]; then
+            local has_marker=false
+            local new_args=()
+            local i=0
+            while [[ $i -lt ${#pytest_args[@]} ]]; do
+                if [[ "${pytest_args[$i]}" == "-m" ]]; then
+                    has_marker=true
+                    new_args+=("-m" "(${pytest_args[$((i+1))]}) and not helm")
+                    i=$((i+2))
+                else
+                    new_args+=("${pytest_args[$i]}")
+                    i=$((i+1))
+                fi
+            done
+            if [[ "$has_marker" == "true" ]]; then
+                pytest_args=("${new_args[@]}")
+            else
+                pytest_args+=("-m" "not helm")
+            fi
+            log_info "DEPLOYMENT_MODE=operator: excluding helm suite"
+        fi
     fi
 
     # Add any extra arguments
--- a/scripts/run-pytest.sh
+++ b/scripts/run-pytest.sh
@@ -36,6 +36,7 @@
 #
 # Filter Options:
 #   --smoke             Run only smoke tests (quick validation)
+#   --operator-gate     Run operator acceptance-gate tests (-m operator)
 #   --slow              Include slow tests (processing, recommendations)
 #
 # Setup Options:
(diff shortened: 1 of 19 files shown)


tests/suites/infrastructure/test_preflight.py
```diff
@@ -17,6 +17,7 @@
 @pytest.mark.infrastructure
 @pytest.mark.component
 @pytest.mark.smoke
+@pytest.mark.operator
```
chadcrum, line 20: This class-level `@pytest.mark.operator` pulls in `test_ingress_pod_exists`, which hard-asserts `component=ingress`. Same pattern on `TestE2ESmoke` (`test_all_critical_pods_running` lists ingress as critical).

That conflicts with `docs/development/operator-test-compat.md` (Ingress not deployed yet) and breaks the `--operator-gate` known-green allowlist.

Can we drop the class-level mark here (and on `TestE2ESmoke`) and only mark methods that are actually green without Ingress?

chadcrum (changes requested):
Recommending changes on one allowlist issue for now: class-level `@pytest.mark.operator` is pulling in Ingress hard-asserts, but the gap docs say Ingress isn't deployed by the operator yet. That breaks the `--operator-gate` known-green contract - see the inline on `TestPodHealth`.

tests/suites/auth/test_rbac_gateway.py
```diff
@@ -161,6 +161,7 @@ def gateway_rbac_iam_user_jwt(
 
 @pytest.mark.auth
 @pytest.mark.integration
+@pytest.mark.operator
```
chadcrum, line 164: Docs say RBAC isn't deployed yet (`{release}-rbac-api` / worker), but class-level `@pytest.mark.operator` here (and on `TestRBACSecurityBoundaries`) pulls a large skip-heavy RBAC suite into `--operator-gate`.

Most of these don't hard-fail the gate - they skip via `_check_gateway_reachable` (`/ingress/ready`) or a missing rbac-api pod. That's hollow signal for a known-green allowlist, not a mass red gate from this file.

Related: `TestOrgAdminIdentityPropagation` asserts 200 on `/rbac/v1/access/` and can be a real hard-fail when RBAC isn't there.

Dropping the class-level marks (and only keeping methods that are actually green without rbac-api) would make `--operator-gate` match what's really green on today's operator.

chadcrum (commented):
Follow-up on the allowlist: RBAC class-level `@pytest.mark.operator` marks - see inline on test_rbac_gateway.py.

tests/suites/cost_management/test_processing.py
```diff
@@ -45,6 +46,7 @@ def test_listener_no_critical_errors(self, cluster_config):
 
 @pytest.mark.cost_management
 @pytest.mark.component
+@pytest.mark.operator
```
chadcrum, line 49: Docs call out the celery worker label mismatch: Helm uses `app.kubernetes.io/component=cost-worker`, operator uses `cost-worker-<queue>`.

Class-level `@pytest.mark.operator` here pulls in `test_celery_workers_ready`, which queries `component=cost-worker` and `pytest.skip`s when nothing matches. On operator that skip is expected, so `--operator-gate` gets a hollow pass instead of a real worker health check.

Dropping the class-level mark (or only marking methods that are green without the Helm worker label) would make `--operator-gate` match what's actually verified on today's operator.

chadcrum (commented):
Follow-up: celery worker label mismatch still hollows out part of the operator gate.

scripts/run-pytest.sh
```diff
@@ -422,6 +426,16 @@ main() {
                 pytest_markers+=("smoke")
                 shift
                 ;;
+            --operator-gate)
+                # Allowlisted tests known-green against operator-managed deploys
+                pytest_markers+=("operator")
```
chadcrum, line 431: `--operator-gate` + suite flags OR-widens selection.

This appends `operator` to `pytest_markers`, and those get joined with `or` later - so `--auth --operator-gate` becomes `(auth) or (operator)` (full auth suite plus the allowlist), not the intersection. Pre-existing multi-flag OR behavior; `--operator-gate` is new and inherits the footgun. Docs show it alone and don't warn about combining.

Documenting "use `--operator-gate` alone" (or making the gate exclusive / AND) would keep runs inside the known-green subset.

chadcrum (commented):
Follow-up: --operator-gate OR-widens when combined with suite flags - see inline on run-pytest.sh.

chadcrum (approved):

chadcrum:
Added some review comments about a few confusing issues, but overall l-g-t-m.

chadcrum:
@testetson22 might want to take a look as well

testetson22 (commented):
I'm wondering if it makes more sense to leverage the existing helm marker + additional dynamic skip logic that detects the runtime environment. We could also add temporary markers only for functional gaps the operator hasn't implemented yet (e.g. `requires_ingress`, `requires_rbac_service`). Essentially, the operator test run becomes -m "not helm and not requires_X and not requires Y" and this would simplify naturally as the operator matures.

Ultimately, I'm not the biggest fan of conflating where a test can run with what a test is via test markers. It's really the domain of initial environment configuration and dynamic resolution to manage runtime concerns.

This also lets us run the full non-helm suite against operator in a non-gating pipeline to track implementation progress, which a positive allowlist can't do as cleanly.

The skip logic would be something simple like:
```
def test_ingress_pod_exists(self, cluster_config):
    if cluster_config.is_operator:
        pytest.skip("Ingress not yet deployed by operator")
    assert check_pod_exists(...)
``` 
 at the top of a test def where cluster_config and requirements are resolved during bootstrapping.

**None of this is a blocking concern, but it is something that I'd likely end up needing to change later to clean up the test config.**

